PR #45, opened 2022-10-18
updated all 4 fallout tests to be able to use more than one proxy nodes, added a test with Astra as target and another two tests to work with DSE-4.7 and 4.8

weideng1:


fallout-tests/dse47-zdmproxy-benchmark-nb-vs-noproxy-astra-target.yaml
```diff
@@ -0,0 +1,350 @@
...
+            datacenters:
+              dc_origin:
+                size: {{origin_node_count}}
+                workload: cassandra
+    - name: target
+      node.count: 1
+      provisioner:
+        name: astra
+        properties:
+          service_account: {{target_astra_service_account}}
```
joao-r-reis, line 65: Remove this on all the astra tests so it uses the default account instead.

I couldn't import the yaml as is because of the service account name.

weideng1: good catch. Let me fix that.

joao-r-reis (approved):

PR #67, opened 2022-10-19
Fix flaky tests and fix #69

### LINKED ISSUE #69: Client connection can be closed before proxy returns protocol error
When a protocol error happens (as part of protocol version negotiation or not), the proxy invokes the shutdown mechanism for that connection after sending the response to the write queue but due to #68 the client may see the connection closed without the protocol error actually being sent.

Due to this, the proxy should at least not shutdown the connection when a protocol error happens while #68 is not addressed.

joao-r-reis:
Fix #69 
Fix more flaky tests

### DIFF
--- a/integration-tests/tls_test.go
+++ b/integration-tests/tls_test.go
@@ -1057,36 +1057,36 @@ func testProxyClientTls(t *testing.T, ccmSetup *setup.CcmTestSetup,
 
 	logMessages := buffer.String()
 
-	warnCheckFail := false
-	warnCheckDone := false
-	for _, errWarnExpected := range proxyTlsConfig.errWarningsExpected {
-		warnCheckDone = true
-		if !strings.Contains(logMessages, errWarnExpected) {
-			t.Logf("%v not found in %v", errWarnExpected, logMessages)
-			warnCheckFail = true
+	warningAssertionFailed := false
+	warningExpected := false
+	for _, expectedWarningMsg := range proxyTlsConfig.errWarningsExpected {
+		warningExpected = true
+		if !strings.Contains(logMessages, expectedWarningMsg) {
+			t.Logf("%v not found in %v", expectedWarningMsg, logMessages)
+			warningAssertionFailed = true
 		}
 	}
 
 	if proxyTlsConfig.errExpected {
 		require.NotNil(t, err, "Did not get expected error %s", proxyTlsConfig.errMsgExpected)
-		errCheckFail := false
-		errCheckDone := false
+		errorAssertionFailed := false
+		errorExpected := false
 		if proxyTlsConfig.errMsgExpected != "" {
-			errCheckDone = true
+			errorExpected = true
 			if !strings.Contains(err.Error(), proxyTlsConfig.errMsgExpected) {
-				errCheckFail = true
+				errorAssertionFailed = true
 				t.Logf("%v not found in %v", err.Error(), proxyTlsConfig.errMsgExpected)
 			}
 		}
-		if errCheckDone && warnCheckDone {
-			require.True(t, !errCheckFail || !warnCheckFail) // only 1 check needs to pass in this scenario
-		} else if errCheckDone {
-			require.False(t, errCheckFail)
-		} else if warnCheckDone {
-			require.False(t, warnCheckFail)
+		if errorExpected && warningExpected {
+			require.False(t, errorAssertionFailed && warningAssertionFailed) // only 1 check needs to pass in this scenario
+		} else if errorExpected {
+			require.False(t, errorAssertionFailed)
+		} else if warningExpected {
+			require.False(t, warningAssertionFailed)
 		}
 	} else {
-		require.False(t, warnCheckFail)
+		require.False(t, warningAssertionFailed)
 		require.Nil(t, err, "testClient setup failed: %v", err)
 		// create schema on clusters through the proxy
 		sendRequest(cqlConn, "CREATE KEYSPACE IF NOT EXISTS testks "+
--- a/integration-tests/tls_test.go
+++ b/integration-tests/tls_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/datastax/zdm-proxy/integration-tests/env"
 	"github.com/datastax/zdm-proxy/integration-tests/setup"
 	"github.com/datastax/zdm-proxy/proxy/pkg/config"
+	zerologger "github.com/rs/zerolog/log"
 	log "github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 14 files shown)


integration-tests/tls_test.go
```diff
@@ -1056,16 +1057,36 @@ func testProxyClientTls(t *testing.T, ccmSetup *setup.CcmTestSetup,
 
 	logMessages := buffer.String()
 
+	warnCheckFail := false
+	warnCheckDone := false
```
grighetto, line 1061: The `Done` suffix here confused me for a second, I thought it would be set when the result of some async processing was done, but in reality this value is simply set when a warning is expected. My suggestion would be to name this `warningExpected` instead. The same goes for `errCheckDone` below, `errExpected`.

integration-tests/tls_test.go
```diff
@@ -1056,16 +1057,36 @@ func testProxyClientTls(t *testing.T, ccmSetup *setup.CcmTestSetup,
...
 		if proxyTlsConfig.errMsgExpected != "" {
-			require.True(t, strings.Contains(err.Error(), proxyTlsConfig.errMsgExpected), err.Error())
+			errCheckDone = true
+			if !strings.Contains(err.Error(), proxyTlsConfig.errMsgExpected) {
+				errCheckFail = true
+				t.Logf("%v not found in %v", err.Error(), proxyTlsConfig.errMsgExpected)
+			}
+		}
+		if errCheckDone && warnCheckDone {
+			require.True(t, !errCheckFail || !warnCheckFail) // only 1 check needs to pass in this scenario
```
grighetto, line 1082: Nit: I think this would be clearer given the other requires are False too:

`require.False(t, errCheckFail && warnCheckFail)`

integration-tests/tls_test.go
```diff
@@ -1340,6 +1361,14 @@ func getClientSideVerifyConnectionCallback(rootCAs *x509.CertPool) func(cs tls.C
 			}
 		}
 		_, err := cs.PeerCertificates[0].Verify(opts)
+		if err != nil {
+			// use zerolog to avoid interacting with proxy's log messages that are used for assertions in the test
```
grighetto, line 1365: Just for my understanding, are we suppressing the log messages from a lib to avoid throwing our log assertions off?

joao-r-reis: The goal is to avoid throwing our log assertions off but we aren't suppressing any log messages. We are capturing the log output from the `logrus` library which is what the proxy uses and the test client uses `zerolog`. To avoid messing with the log assertions I decided to use `zerolog` here on this test code in particular instead of `logrus`

integration-tests/tls_test.go
joao-r-reis: I actually set this to `require.True` because to me it felt easier to understand but since I added a comment anyway I can use `require.False` to be consistent with the others

proxy/pkg/zdmproxy/clientconn.go
```diff
@@ -174,14 +175,15 @@ func (cc *ClientConnector) listenForRequests() {
...
 			} else if protocolErrResponseFrame != nil {
-				f = protocolErrResponseFrame
-				if !protocolErrOccurred {
-					protocolErrOccurred = true
-					cc.sendResponseToClient(protocolErrResponseFrame)
-					cc.clientHandlerShutdownRequestCancelFn()
-					setDrainModeNowFunc()
-					continue
-				}
+				alreadySentProtocolErr = protocolErrResponseFrame
```
alicel, line 178: should this assignment happen after `cc.sendResponseToClient(protocolErrResponseFrame)`?

proxy/pkg/zdmproxy/clientconn.go
```diff
@@ -174,14 +175,15 @@ func (cc *ClientConnector) listenForRequests() {
...
+				protocolErrOccurred = true
+				cc.sendResponseToClient(protocolErrResponseFrame)
+				continue
+			} else if alreadySentProtocolErr != nil {
```
alicel, line 182: Just to check that I understand correctly. The intended behaviour with this change is: if a protocol error is sent on the connection, then we want to ensure that all other potentially pending responses that may still be returned are replaced by that protocol error (unless they were themselves protocol errors in the first place). So we're keeping the already returned protocol error to replace the frame of any subsequent non-protocol error response with it.

The idea is to ensure that we do not send seemingly valid responses on a connection on which a protocol error was returned while this connection is being drained. 

Is this correct?

integration-tests/tls_test.go
joao-r-reis: done

integration-tests/tls_test.go
joao-r-reis: done

proxy/pkg/zdmproxy/clientconn.go
```diff
@@ -174,14 +175,15 @@ func (cc *ClientConnector) listenForRequests() {
 					err, cc.clientHandlerContext, cc.clientHandlerCancelFunc, ClientConnectorLogPrefix, "reading", connectionAddr)
 				break
 			} else if protocolErrResponseFrame != nil {
-				f = protocolErrResponseFrame
-				if !protocolErrOccurred {
-					protocolErrOccurred = true
-					cc.sendResponseToClient(protocolErrResponseFrame)
-					cc.clientHandlerShutdownRequestCancelFn()
```
grighetto, line 181: It looks like `clientHandlerShutdownRequestCancelFn` is not called anywhere anymore, we should probably remove it from the struct.

alicel (approved):
Looks fine to me. I left a question and a tentative suggestion but nothing that would prevent it from being approved.

proxy/pkg/zdmproxy/clientconn.go
joao-r-reis: Yes, but also the main change is that we no longer trigger the connection shutdown mechanism so the connection will be left in an unusable state (constant protocol errors) but it will be up to the driver to close it (at least until we address #68 )

proxy/pkg/zdmproxy/clientconn.go
joao-r-reis: Doesn't really matter because there is no concurrency here, it's a single goroutine with a `for` loop

proxy/pkg/zdmproxy/clientconn.go
joao-r-reis: We want to call it once #68 is addressed, maybe I can add a comment about this

joao-r-reis: done

grighetto (approved):
👍

PR #75, opened 2022-11-10
Fix request deadlock when a lot of connections are opened concurrently

### LINKED ISSUE #74: Handshakes with auth enabled can deadlock if multiple handshakes are happening concurrently
In `handleHandshakeRequest` there is a blocking channel read operation happening to wait until the secondary handshake is done. This blocking read waits for a response within a scheduler that is also used for response processing so a deadlock can occur. 

If auth is disabled then no requests will be sent so no wait will be performed and no deadlock will occur.

┆Issue is synchronized with this [Jira Bug](https://datastax.jira.com/browse/ZDM-486) by [Unito](https://www.unito.io)
┆Components: Proxy
┆Issue Number: ZDM-486
┆Link To Issue: https://datastax.jira.com/browse/ZDM-486

joao-r-reis:
Fix #74

### DIFF
--- a/integration-tests/stress_test.go
+++ b/integration-tests/stress_test.go
@@ -1,0 +1,172 @@
+package integration_tests
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"github.com/gocql/gocql"
+	"github.com/riptano/cloud-gate/integration-tests/env"
+	"github.com/riptano/cloud-gate/integration-tests/setup"
+	"github.com/rs/zerolog"
+	log "github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+)
+
+func TestSimultaneousConnections(t *testing.T) {
+	if !env.UseCcm {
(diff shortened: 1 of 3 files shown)


weideng1:
I'll leave the code review for other folks, but just a quick note: I've run a docker image of this PR (https://hub.docker.com/layers/datastax/zdm-proxy/sha-8f852d4/images/sha256-81754cf794af5e22edd76cfecc8b8aacc44eb1da7d8bdeb8c494ab7ec71ed44f?context=explore) in my k8s test environment with various proxy pod restart conditions and increased connection counts, and so far haven't run into any issue yet.

integration-tests/stress_test.go
```diff
@@ -0,0 +1,172 @@
...
+	"github.com/stretchr/testify/require"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+)
+
+func TestSimultaneousConnections(t *testing.T) {
+	if !env.RunCcmTests {
+		t.Skip("Test requires CCM, set USE_CCM env variable to TRUE")
```
alicel, line 22: This message should say `Test requires CCM, set RUN_CCMTESTS env variable to TRUE`.

alicel (approved):
LGTM. Left a couple of minor comments on the test code.

PR #78, opened 2022-11-21
ZDM-287 Send heartbeats to clusters

grighetto:
@joao-r-reis There's still a couple of things in flux right now (I'll comment directly in the PR files), but I'd like to request your feedback. This should be mostly in line with the approach we discussed initially, the stream id mapping idea, but it's nothing set in stone, we can still change it as needed.

I've tested it locally with the integration tests and the nosqlbench test in docker compose.


proxy/pkg/zdmproxy/clusterconn.go
```diff
@@ -472,3 +477,20 @@ func (cc *ClusterConnector) sendAsyncRequest(
...
+
+var heartBeatTime = time.Now()
+
+func (cc *ClusterConnector) sendHeartbeat() {
+	if time.Since(heartBeatTime) < 1*time.Minute {
+		return
+	}
+	heartBeatTime = time.Now()
+	optionsMsg := &message.Options{}
+	heartBeatFrame := frame.NewFrame(ccProtocolVersion, -1, optionsMsg)
```
grighetto, line 489: Using the static `ccProtocolVersion` const here causes some errors in the logs. It points to v3, but if the proxy has negotiated v4, it will complain.
@joao-r-reis Can I determine the version in use from here?

joao-r-reis: `ccProtocolVersion` should be used on the control connection only. 
I believe the best way to get the protocol version here is to look at the startup frame that is stored on the client handler (`ch.startupRequest`) but in order to do that you will have to change that field to be thread safe because currently it is not.

joao-r-reis (commented):
Looking good! I left some comments.

Also I think it would be great if you could add a node metric for the number of used stream ids. It could be easily implemented by a gauge function metric that checks the result from `len(ids)`. It would be a metric for each cluster connector including the async connector.

alicel (commented):
Looks good overall. Made some suggestions but they are all minor things.

proxy/pkg/zdmproxy/coalescer.go
```diff
@@ -137,7 +141,12 @@ func (recv *writeCoalescer) RunWriteQueueLoop() {
 					}
 
 					log.Tracef("[%v] Writing %v on %v", recv.logPrefix, f.Header, connectionAddr)
-					err := writeRawFrame(tempBuffer, connectionAddr, recv.shutdownContext, f)
+					var newf = f
+					if recv.frameProcessor != nil {
+						newf = f.Clone()
+						recv.frameProcessor.AssignUniqueId(newf)
+					}
+					err := writeRawFrame(tempBuffer, connectionAddr, recv.shutdownContext, newf)
```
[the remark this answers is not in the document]
grighetto: I've tried that initially but the tests fail if I don't clone the frame, I haven't determined exactly why yet, but I imagine it's because there are references being held somewhere else.

proxy/pkg/zdmproxy/clusterconn.go
```diff
@@ -472,3 +477,20 @@ func (cc *ClusterConnector) sendAsyncRequest(
 
 	return err == nil
 }
+
+var heartBeatTime = time.Now()
+
+func (cc *ClusterConnector) sendHeartbeat() {
+	if time.Since(heartBeatTime) < 1*time.Minute {
```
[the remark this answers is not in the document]
grighetto: Done

proxy/pkg/zdmproxy/frameprocessor.go
```diff
@@ -0,0 +1,97 @@
...
+	frame.Header.StreamId = newId
+	return nil
+}
+
+func (sip *internalCqlStreamIdProcessor) ReleaseId(frame *frame.Frame) {
+	if frame == nil {
+		return
+	}
+	sip.mapper.ReleaseId(frame.Header.StreamId)
+}
```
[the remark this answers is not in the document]
grighetto: My main motivation for having separate interfaces was to prevent someone from inadvertently passing the wrong implementation of the `StreamIdMapper` to `CqlConnection`. There would be no compile time error, but the runtime behavior would be broken. It's just a way to enforce the more specialized implementation.

proxy/pkg/zdmproxy/pendingrequests.go
```diff
@@ -42,13 +36,11 @@ func (p *pendingRequests) getOrCreateRequestContextHolder(streamId int16) *reque
...
-	streamId, err := p.reserveStreamId()
-	if err != nil {
-		return -1, fmt.Errorf("stream id map ran out of stream ids: %w", err)
-	}
+func (p *pendingRequests) store(reqCtx RequestContext, frame *frame.RawFrame) (int16, error) {
+	p.frameProcessor.AssignUniqueId(frame)
+	streamId := frame.Header.StreamId
 	holder := getOrCreateRequestContextHolder(p.pending, streamId)
-	err = holder.SetIfEmpty(reqCtx)
+	var err = holder.SetIfEmpty(reqCtx)
```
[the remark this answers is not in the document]
grighetto: Fixed

proxy/pkg/zdmproxy/streamidmapper.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+type cqlStreamIdMapper struct {
+	sync.Mutex
+	ids chan int16
+}
+
+func NewCqlStreamIdMapper() InternalCqlStreamIdMapper {
+	var ids = make(chan int16, maxStreamIds)
+	for i := int16(0); i < maxStreamIds; i++ {
```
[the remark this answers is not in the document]
grighetto: I took a stab at implementing this. The max stream ids is now exposed as a config var. Do you think we need anything more for this PR (and soon release)? We can probably tackle the v2 support in a follow-on work. Wdyt?

proxy/pkg/zdmproxy/streamidmapper.go
```diff
@@ -0,0 +1,107 @@
...
+	syntheticId = <-sim.clusterIds
+	sim.idMapper[streamId] = syntheticId
+	sim.synMapper[syntheticId] = streamId
+	return syntheticId, nil
+}
+
+func (sim *streamIdMapper) RestoreId(syntheticId int16) (int16, error) {
+	sim.Lock()
+	defer sim.Unlock()
+	syntheticId, contains := sim.synMapper[syntheticId]
```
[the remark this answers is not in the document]
grighetto: Thanks for spotting that, fixed the variable name.

proxy/pkg/zdmproxy/frameprocessor.go
```diff
@@ -0,0 +1,97 @@
+package zdmproxy
+
+import "github.com/datastax/go-cassandra-native-protocol/frame"
+
+// FrameProcessor manages the mapping of incoming stream ids to the actual ids that are sent over to the clusters.
+// This is important to prevent overlapping ids of client-side requests and proxy internal requests (such as heartbeats)
+// that are written through the same connection to the cluster.
+type FrameProcessor interface {
```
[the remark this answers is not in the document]
grighetto: I went back and forth with the names a few times to be honest. My original goal was indeed to have a general concept where we could have multiple `FrameProcessor` implementations to mutate the frames before and after sending/receiving them (actually, those were the method names I had initially, `FrameProcessor.Before` and `FrameProcessor.After`. But then I realized we might not always mutate the frames at the same point in the code, it depends on what we want to do, if it's coming from the client, if it's being generated internally, etc.

In summary, I agree we can probably change the FrameProcessor name to something more specific, but I still would like to keep them as separate interfaces. The StreamIdMapper only cares about ints, it doesn't have to know about Frames or RawFrames, that's good for separation of concerns and also for testing. Wtdy?

proxy/pkg/zdmproxy/clusterconn.go
```diff
@@ -472,3 +477,20 @@ func (cc *ClusterConnector) sendAsyncRequest(
 
 	return err == nil
 }
+
+var heartBeatTime = time.Now()
```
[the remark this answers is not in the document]
grighetto: Thanks for pointing that out. I actually moved this var into the clusterconn struct, which was probably the right place from the start.

proxy/pkg/zdmproxy/clusterconn.go
grighetto: Thanks for the pointer, that seems to work in my initial tests (pushed the change to use the startupRequest in the `sendHeartbeat` function). I still need to add the mutex around that var though, that's coming next.

proxy/pkg/zdmproxy/coalescer.go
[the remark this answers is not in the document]
joao-r-reis: I think it would be wise to understand where those failures are coming from, this `clone` call might just be hiding the issue.

proxy/pkg/zdmproxy/frameprocessor.go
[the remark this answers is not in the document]
joao-r-reis: Unit/integration tests are there to prevent that, in some cases I like having compile time safeguards but it's very easy to have unit tests covering this here so I don't think this extra code complexity is worth it.

proxy/pkg/zdmproxy/clienthandler.go
```diff
@@ -1434,10 +1435,12 @@ func (ch *ClientHandler) executeRequest(
 		log.Tracef("Forwarding request with opcode %v for stream %v to %v",
 			f.Header.OpCode, f.Header.StreamId, common.ClusterTypeOrigin)
 		ch.originCassandraConnector.sendRequestToCluster(originRequest)
+		ch.targetCassandraConnector.sendHeartbeat(ch.startupRequest.Header.Version, ch.conf.HeartbeatIntervalMs)
 	case forwardToTarget:
 		log.Tracef("Forwarding request with opcode %v for stream %v to %v",
 			f.Header.OpCode, f.Header.StreamId, common.ClusterTypeTarget)
 		ch.targetCassandraConnector.sendRequestToCluster(targetRequest)
+		ch.originCassandraConnector.sendHeartbeat(ch.startupRequest.Header.Version, ch.conf.HeartbeatIntervalMs)
```
joao-r-reis, line 1443: just leaving a note here so I don't forget to check this in a future review:

make startup threadsafe (doesn't have to be a mutex, an atomic value is probably fine)

proxy/pkg/zdmproxy/coalescer.go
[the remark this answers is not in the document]
joao-r-reis: Actually, it might be a good idea to try and create a function that clones only the header part of the frame but doesn't clone the body and the frame processor should call that function as well as here.

proxy/pkg/zdmproxy/streamidmapper.go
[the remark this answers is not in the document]
joao-r-reis: I think it's ok to leave it like this for now

proxy/pkg/zdmproxy/clusterconn.go
grighetto: Turned `startupRequest` into an atomic value.

proxy/pkg/zdmproxy/coalescer.go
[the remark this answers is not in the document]
grighetto: @joao-r-reis Yeah, I've been eager to investigate this deeper but I wanted to get the low-hanging fruits out of the way first.
I think I got to the bottom of this now. The problem is that the same `*frame.RawFrame` reference is passed to both origin and target cluster connectors, so with *no cloning* when the frameprocessor changes the frame's stream id, it changes it in both places.
I tried cloning just the Header struct as you suggested, but when we do `frame.Header = clonedHeader`, it will still change the reference in both places (because the parent is the same).
In the end I had to create a new parent `frame.RawFrame` struct and pass it the original Body with the cloned Header.

proxy/pkg/zdmproxy/frameprocessor.go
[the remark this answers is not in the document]
grighetto: @joao-r-reis Alright, I've consolidated the StreamIdMapper interfaces into a single one.
I agree we can do those checks in tests, but it's arguably easier to understand a clear API contract than figuring out why a test broke after the fact. I was just trying to advocate for both.

proxy/pkg/zdmproxy/streamidmapper.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	return &cqlStreamIdMapper{
+		ids: ids,
+	}
+}
+
+func (icsim *cqlStreamIdMapper) GetNewId() (int16, error) {
+	icsim.Lock()
+	defer icsim.Unlock()
+	return <-icsim.ids, nil
```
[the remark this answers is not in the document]
grighetto: For now, I added the select/case construct to avoid blocking and also logged an error message.

proxy/pkg/zdmproxy/streamidmapper.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func (icsim *cqlStreamIdMapper) ReleaseId(syntheticId int16) {
+	icsim.Lock()
+	defer icsim.Unlock()
+	icsim.ids <- syntheticId
```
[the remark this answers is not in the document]
grighetto: For now, I added the select/case construct to avoid blocking and also logged an error message.

proxy/pkg/zdmproxy/clienthandler.go
grighetto: @joao-r-reis Alright, I turned the `startupRequest` var into an atomic value.

proxy/pkg/zdmproxy/clusterconn.go
```diff
@@ -472,3 +480,18 @@ func (cc *ClusterConnector) sendAsyncRequest(
 
 	return err == nil
 }
+
+func (cc *ClusterConnector) sendHeartbeat(version primitive.ProtocolVersion, heartbeatIntervalMs int) {
+	if time.Since(cc.lastHeartbeatTime) < time.Duration(heartbeatIntervalMs)*time.Millisecond {
+		return
+	}
+	cc.lastHeartbeatTime = time.Now()
```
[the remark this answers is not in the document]
grighetto: @joao-r-reis Good call. I implemented the double-check locking per your suggestion.

grighetto:
> Also I think it would be great if you could add a node metric for the number of used stream ids. It could be easily implemented by a gauge function metric that checks the result from len(ids). It would be a metric for each cluster connector including the async connector.

@joao-r-reis I've added the gauge metrics for all the connectors. Can you take a look?

grighetto:
@alicel @joao-r-reis I believe I've addressed all the remaining items, this is ready for another pass. Please take a look when you have the chance.

proxy/pkg/zdmproxy/streamidmapper.go
```diff
@@ -0,0 +1,130 @@
...
+	return -1, fmt.Errorf("internal CQL ids have no original ids to be restored")
+}
+
+func (icsim *cqlStreamIdMapper) ReleaseId(syntheticId int16) (int16, error) {
+	icsim.Lock()
+	defer icsim.Unlock()
+	select {
+	case icsim.ids <- syntheticId:
+	default:
+		log.Errorf("stream ids channel full, ignoring id %v", syntheticId)
```
joao-r-reis, line 60: shouldn't this return the err instead of logging it?

proxy/pkg/zdmproxy/streamidmapper.go
```diff
@@ -0,0 +1,130 @@
...
+	defer sim.Unlock()
+	var originalId = sim.synMapper[syntheticId]
+	delete(sim.idMapper, originalId)
+	delete(sim.synMapper, syntheticId)
+	select {
+	case sim.clusterIds <- syntheticId:
+	default:
+		log.Errorf("stream ids channel full, ignoring id %v", syntheticId)
+	}
+	return originalId, nil
```
joao-r-reis, line 129: I'm struggling to understand the difference between `ReleaseId` and `RestoreId` in practice. In the POV of the stream id mapper, it shouldn't matter if we are restoring or releasing, in practice the mapper needs to release the ID, delete it from the maps and return the original one. Whether this is being called because the request timed out or because a response was returned by the server shouldn't matter.

proxy/pkg/zdmproxy/streamidmapper.go
```diff
@@ -0,0 +1,130 @@
...
+
+func (icsim *cqlStreamIdMapper) ReleaseId(syntheticId int16) (int16, error) {
+	icsim.Lock()
+	defer icsim.Unlock()
+	select {
+	case icsim.ids <- syntheticId:
+	default:
+		log.Errorf("stream ids channel full, ignoring id %v", syntheticId)
+	}
+	return -1, nil
```
joao-r-reis, line 62: This should always return the `syntheticId` because this mapper is essentially not doing any mapping but from the outside it can be used as if it was mapping (which is kind of the contract of the interface)

proxy/pkg/zdmproxy/streamidmapper.go
```diff
@@ -0,0 +1,130 @@
...
+		} else {
+			return -1, fmt.Errorf("stream id channel closed")
+		}
+	default:
+		return -1, fmt.Errorf("no stream id available")
+	}
+}
+
+func (icsim *cqlStreamIdMapper) RestoreId(_ int16) (int16, error) {
+	return -1, fmt.Errorf("internal CQL ids have no original ids to be restored")
```
joao-r-reis, line 51: This should always return the original id even though this mapper is essentially not doing any mapping because from the outside it can be used as if it was mapping (which is kind of the contract of the interface)

proxy/pkg/zdmproxy/streamidmapper.go
```diff
@@ -0,0 +1,130 @@
...
+		ids <- i
+	}
+	return &cqlStreamIdMapper{
+		ids: ids,
+	}
+}
+
+func (icsim *cqlStreamIdMapper) GetNewIdFor(_ int16) (int16, error) {
+	icsim.Lock()
+	defer icsim.Unlock()
```
joao-r-reis, line 37: Why do we need locking? go channels are thread safe

proxy/pkg/zdmproxy/streamidmapper.go
```diff
@@ -0,0 +1,130 @@
...
+	return &streamIdMapper{
+		idMapper:   idMapper,
+		clusterIds: streamIdsQueue,
+		synMapper:  synMapper,
+	}
+}
+
+func (sim *streamIdMapper) GetNewIdFor(streamId int16) (int16, error) {
+	sim.Lock()
+	defer sim.Unlock()
```
joao-r-reis, line 88: We should try to reduce locking in this, I think we can use the `sync.Map` instead of regular map and instead of removing entries from the map we can set them to nil or an empty global object (this is because `sync.Map` does not perform well with deletes).

An alternative is to just move the locking to the part where the maps are being updated (possibly remove the `contains` check altogether - check my comment below) and benchmark to see if it's good enough.

proxy/pkg/zdmproxy/frameprocessor.go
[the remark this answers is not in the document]
joao-r-reis: I think the frame processor interfaces should be merged as well, same reasoning. 

Also I mentioned this in another comment but the stream id's `RestoreId` and `ReleaseId` can probably be merged into a single method because the stream id mapper doesn't have to know if a stream id is being released or restored, it's essentially the same process as the mapper is concerned. And the internal frame processor's implementation of `Restore` can just return the same frame.

proxy/pkg/zdmproxy/coalescer.go
```diff
@@ -137,6 +141,13 @@ func (recv *writeCoalescer) RunWriteQueueLoop() {
 					}
 
 					log.Tracef("[%v] Writing %v on %v", recv.logPrefix, f.Header, connectionAddr)
+					if recv.frameProcessor != nil {
+						f = &frame.RawFrame{
+							Header: f.Header,
+							Body:   f.Body,
+						}
```
joao-r-reis, line 148: I think it would be less error prone to move this to the frame processor itself and make the frame processor return an actual frame (the internal processor would return the same frame while the connector's frame processor would return a new frame with the same body but cloned header)

joao-r-reis (commented):
Second pass, Functionally I think it's looking pretty good but I still added some comments related to code structure, metrics and locking.

proxy/pkg/zdmproxy/frameprocessor.go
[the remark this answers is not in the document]
joao-r-reis: I think the `StreamIdProcessor` interface should be changed to `StreamIdMapper`

proxy/pkg/metrics/proxy_metrics.go
```diff
@@ -5,6 +5,9 @@ const (
 	typeReadsTarget = "reads_target"
 	typeWrites      = "writes"
 	typeCount       = "count"
+	typeIdsOrigin   = "ids_origin"
+	typeIdsTarget   = "ids_target"
+	typeIdsAsync    = "ids_async"
```
joao-r-reis, line 10: I would remove the `ids_` part, `connector=origin` makes more sense than `connector=ids_origin`

proxy/pkg/zdmproxy/streamidmapper.go
```diff
@@ -0,0 +1,62 @@
...
+	select {
+	case id, ok := <-sim.clusterIds:
+		if ok {
+			sim.idMapper[id] = streamId
```
joao-r-reis, line 41: We should check if there already is an entry for this id because if there is then it's a stream id collision (probably a bug) and we need to return an error instead. 

I think I originally mentioned we could maybe remove the check but I think it does make sense to check it. If we had the check I think you would have noticed a bug in the control connection.

proxy/pkg/zdmproxy/streamidmapper.go
joao-r-reis: Update: I think we can just move the locking to the actual map operation part (after reading from the channel) and then we can see how it performs before thinking about optimizations

joao-r-reis (commented):
Looking good! I've left another batch of comments. I'll go through the older ones and mark them as resolved where it makes sense.

integration-tests/metrics_test.go
```diff
@@ -62,6 +62,10 @@ var proxyMetrics = []metrics.Metric{
 	metrics.InFlightWrites,
 
 	metrics.OpenClientConnections,
+
+	metrics.UsedStreamIdsOrigin,
+	metrics.UsedStreamIdsTarget,
+	metrics.UsedStreamIdsAsync,
```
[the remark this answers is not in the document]
joao-r-reis: This one is marked as "outdated" because of the rename but it's still unresolved

proxy/pkg/metrics/proxy_metrics.go
```diff
@@ -141,4 +185,9 @@ type ProxyMetrics struct {
 	InFlightWrites      Gauge
 
 	OpenClientConnections GaugeFunc
+
+	ProxyUsedStreamIdsOrigin  Gauge
```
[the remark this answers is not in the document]
joao-r-reis: This is marked as "outdated" but it's still unresolved. I think it is extra relevant now that the metrics were renamed to look like the metric is tracking the "available" stream ids (at least that's how I interpret it).
With the latest changes I think this metric should be changed to a gaugefunc that returns `len(streamIdQueue)`

proxy/pkg/zdmproxy/clusterconn.go
```diff
@@ -236,6 +249,7 @@ func (cc *ClusterConnector) runResponseListeningLoop() {
 		protocolErrOccurred := false
 		for {
 			response, err := readRawFrame(bufferedReader, connectionAddr, cc.clusterConnContext)
+			cc.frameProcessor.RestoreId(response)
```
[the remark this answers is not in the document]
joao-r-reis: restoreid checks if response is nil so this is fine.

proxy/pkg/zdmproxy/coalescer.go
joao-r-reis: I think I already commented something similar to this on the setstreamid methods but basically I think this should be moved to those 2 methods and then the new frame should be returned.

joao-r-reis: And the stream id mapper implementation for the control connection doesn't have to clone anything imo, it can just return the existing frame.

proxy/pkg/zdmproxy/streamidmapper.go
grighetto: Alright, I've added the check back.

proxy/pkg/zdmproxy/cqlconn.go
```diff
@@ -263,25 +258,17 @@ func (c *cqlConn) sendContext(request *frame.Frame, ctx context.Context) (chan *
...
-	var streamId int16
-	select {
-	case streamId = <-c.streamIdQueue:
-		respChan = make(chan *frame.Frame, 1)
-	default:
-		return nil, fmt.Errorf("no available stream ids for request")
-	}
+	var respChan = make(chan *frame.Frame, 1)
+
 	c.pendingOperationsLock.Lock()
```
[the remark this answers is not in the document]
grighetto: Done

proxy/pkg/zdmproxy/cqlconn.go
```diff
@@ -164,7 +161,6 @@ func (c *cqlConn) StartResponseLoop() {
 			}
 
 			delete(c.pendingOperations, f.Header.StreamId)
-			c.streamIdQueue <- f.Header.StreamId
 			c.pendingOperationsLock.Unlock()
```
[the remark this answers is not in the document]
grighetto: Done

proxy/pkg/zdmproxy/frameprocessor.go
```diff
@@ -62,61 +42,64 @@ func (sip *streamIdProcessor) AssignUniqueId(frame *frame.RawFrame) error {
...
 	newHeader := f.Header.Clone()
 	newHeader.StreamId = id
 	f.Header = newHeader
 }
 
-func setRawFrameStreamId(f *frame.Frame, id int16) {
+func setFrameStreamId(f *frame.Frame, id int16) {
 	newHeader := f.Header.Clone()
 	newHeader.StreamId = id
 	f.Header = newHeader
```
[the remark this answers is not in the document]
grighetto: Done

proxy/pkg/zdmproxy/frameprocessor.go
```diff
@@ -11,15 +11,9 @@ import (
 // that are written through the same connection to the cluster.
 type FrameProcessor interface {
 	AssignUniqueId(frame *frame.RawFrame) error
```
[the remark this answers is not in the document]
grighetto: Done

joao-r-reis (commented):
Looking good, a couple more comments and also I think we should have a simulacron test with multiple test cases to verify that only the secondary is getting heartbeats from the proxy and also the number of heartbeats in a time interval matches our expectations.

proxy/pkg/zdmproxy/clienthandler.go
```diff
@@ -1423,6 +1425,12 @@ func (ch *ClientHandler) executeRequest(
 		reqCtx.SetTimer(timer)
 	}
 
+	startupFrameInterface := ch.startupRequest.Load()
+	var startupFrameVersion primitive.ProtocolVersion
+	if startupFrameInterface != nil {
+		startupFrameVersion = startupFrameInterface.(*frame.RawFrame).Header.Version
+	}
```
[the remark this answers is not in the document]
grighetto: The `sendHeartbeat` function in `clusterconn.go` will now check if the version param is different from zero (which will be the case when the `startupFrameFrameInterface` is nil) before sending the beat.

proxy/pkg/metrics/proxy_metrics.go
grighetto: Done

proxy/pkg/zdmproxy/frameprocessor.go
```diff
@@ -62,61 +42,64 @@ func (sip *streamIdProcessor) AssignUniqueId(frame *frame.RawFrame) error {
 	if err != nil {
 		return err
 	}
-	setFrameStreamId(frame, newId)
+	setRawFrameStreamId(frame, newId)
 	sip.metrics.Add(1)
```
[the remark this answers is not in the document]
grighetto: Added missing nil check just like in the other functions.

proxy/pkg/zdmproxy/clusterconn.go
```diff
@@ -472,3 +498,30 @@ func (cc *ClusterConnector) sendAsyncRequest(
...
+	defer cc.lastHeartbeatLock.Unlock()
+	if !cc.shouldSendHeartbeat(heartbeatIntervalMs) {
+		return
+	}
+	cc.lastHeartbeatTime.Store(time.Now())
+	optionsMsg := &message.Options{}
+	heartBeatFrame := frame.NewFrame(version, -1, optionsMsg)
+	rawFrame, err := defaultCodec.ConvertToRawFrame(heartBeatFrame)
+	if err != nil {
+		log.Error("Cannot convert heartbeat frame to raw frame")
```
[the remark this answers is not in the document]
grighetto: Done

proxy/pkg/zdmproxy/coalescer.go
```diff
@@ -136,8 +140,14 @@ func (recv *writeCoalescer) RunWriteQueueLoop() {
 						ok = true
 					}
 
+					var err error = nil
```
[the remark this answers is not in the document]
grighetto: Done

proxy/pkg/zdmproxy/cqlconn.go
```diff
@@ -164,7 +161,10 @@ func (c *cqlConn) StartResponseLoop() {
 			}
 
 			delete(c.pendingOperations, f.Header.StreamId)
-			c.streamIdQueue <- f.Header.StreamId
+			f, err = c.frameProcessor.ReleaseIdFrame(f)
+			if err != nil {
+				log.Error(err)
```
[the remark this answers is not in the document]
grighetto: Done - added prefix to the message.

proxy/pkg/zdmproxy/streamidmapper.go
```diff
@@ -0,0 +1,64 @@
...
+		}
+	default:
+		return -1, fmt.Errorf("no stream id available")
+	}
+}
+
+func (sim *streamIdMapper) ReleaseId(syntheticId int16) (int16, error) {
+	sim.Lock()
+	defer sim.Unlock()
+	var originalId = sim.idMapper[syntheticId]
```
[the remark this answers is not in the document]
grighetto: Ok, I added the check, the problem then becomes what to return for the original stream id in that case.

proxy/pkg/zdmproxy/frameprocessor.go
```diff
@@ -0,0 +1,92 @@
...
+
+func (sip *streamIdProcessor) ReleaseIdFrame(frame *frame.Frame) (*frame.Frame, error) {
+	if frame == nil {
+		return frame, nil
+	}
+	var _, err = sip.mapper.ReleaseId(frame.Header.StreamId)
+	if sip.metrics != nil && err != nil {
+		sip.metrics.Subtract(1)
+	}
+	return frame, err
```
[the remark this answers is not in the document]
grighetto: Alright, in the beginning of this PR I had two different mapper implementations and one of them had only the channel for providing the ids for internal requests as you're describing here. I did change that again to match your latest proposal, so this is done.

proxy/pkg/zdmproxy/coalescer.go
```diff
@@ -136,8 +140,14 @@ func (recv *writeCoalescer) RunWriteQueueLoop() {
...
 					log.Tracef("[%v] Writing %v on %v", recv.logPrefix, f.Header, connectionAddr)
-					err := writeRawFrame(tempBuffer, connectionAddr, recv.shutdownContext, f)
+					if recv.frameProcessor != nil {
+						f, err = recv.frameProcessor.AssignUniqueId(f)
```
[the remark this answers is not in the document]
grighetto: Done

proxy/pkg/zdmproxy/streamidmapper.go
grighetto: > Update: I think we can just move the locking to the actual map operation part (after reading from the channel) and then we can see how it performs before thinking about optimizations

Done

proxy/pkg/zdmproxy/frameprocessor.go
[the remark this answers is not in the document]
joao-r-reis: Yes my proposal from the start was to unify the mapper interface but not the implementation. The processor however I did propose to unify both the implementation and the interface because the differentiation could be made at mapper level.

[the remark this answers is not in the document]
grighetto: Gotcha, I think we're on the same page now.

proxy/pkg/zdmproxy/clusterconn.go
```diff
@@ -459,16 +496,43 @@ func (cc *ClusterConnector) sendAsyncRequest(
...
+	defer cc.lastHeartbeatLock.Unlock()
+	if !cc.shouldSendHeartbeat(heartbeatIntervalMs) {
+		return
+	}
+	cc.lastHeartbeatTime.Store(time.Now())
+	optionsMsg := &message.Options{}
+	heartBeatFrame := frame.NewFrame(version, -1, optionsMsg)
+	rawFrame, err := defaultCodec.ConvertToRawFrame(heartBeatFrame)
+	if err != nil {
+		log.Errorf("Cannot convert heartbeat frame to raw frame: %v", err)
```
joao-r-reis, line 527: missing a return here

proxy/pkg/zdmproxy/clusterconn.go
```diff
@@ -432,20 +466,23 @@ func (cc *ClusterConnector) sendAsyncRequest(
 	if !cc.validateAsyncStateForRequest(asyncRequest) {
 		return false
 	}
-
 	asyncReqCtx := NewAsyncRequestContext(requestInfo, asyncRequest.Header.StreamId, expectedResponse, overallRequestStartTime)
-	var newStreamId int16
-	newStreamId, err := cc.asyncPendingRequests.store(asyncReqCtx)
+
+	var err error = nil
```
joao-r-reis, line 471: var err error

proxy/pkg/zdmproxy/clusterconn.go
```diff
@@ -361,7 +390,16 @@ func (cc *ClusterConnector) handleAsyncResponse(response *frame.RawFrame) *frame
...
 
 func (cc *ClusterConnector) sendRequestToCluster(frame *frame.RawFrame) {
-	cc.writeCoalescer.Enqueue(frame)
+	var err error
+	if cc.frameProcessor != nil {
+		frame, err = cc.frameProcessor.AssignUniqueId(frame)
+	}
+	if err != nil {
+		log.Errorf("couldn't assign unique id to frame %v", frame)
+		handleConnectionError(err, cc.clusterConnContext, cc.cancelFunc, string(cc.connectorType), "writing", cc.connection.RemoteAddr().String())
```
joao-r-reis, line 399: hmm I'm not sure if calling `handleConnectionError` is the right choice here, this function is usually called by the read/write goroutines when an IO error happens to trigger a shutdown operation on all connections related to this connector (client and other cluster).

In this scenario I think it's better to just log the error and return

proxy/pkg/zdmproxy/clusterconn.go
```diff
@@ -361,7 +390,16 @@ func (cc *ClusterConnector) handleAsyncResponse(response *frame.RawFrame) *frame
 }
 
 func (cc *ClusterConnector) sendRequestToCluster(frame *frame.RawFrame) {
-	cc.writeCoalescer.Enqueue(frame)
+	var err error
+	if cc.frameProcessor != nil {
+		frame, err = cc.frameProcessor.AssignUniqueId(frame)
+	}
+	if err != nil {
+		log.Errorf("couldn't assign unique id to frame %v", frame)
```
joao-r-reis, line 398: `log.Errorf("[%v] Couldn't assign stream id to frame %v: %v, frame.Header.OpCode, err)`

proxy/pkg/zdmproxy/streamidmapper.go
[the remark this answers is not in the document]
joao-r-reis: I think every time there's an error in these methods the returned value should be `-1`, the caller shouldn't rely on it if an error has occured.

proxy/pkg/zdmproxy/streamidmapper.go
```diff
@@ -0,0 +1,107 @@
...
+		return -1, fmt.Errorf("no stream id available")
+	}
+}
+
+func (sim *streamIdMapper) ReleaseId(syntheticId int16) (int16, error) {
+	sim.Lock()
+	originalId, contains := sim.idMapper[syntheticId]
+	if !contains {
+		sim.Unlock()
+		return originalId, fmt.Errorf("trying to release a stream id not found in mapper: %v", syntheticId)
```
joao-r-reis, line 97: `-1`

joao-r-reis (commented):
This pass resulted in comments that are mostly about small stuff, the PR is in a very good place now and I think it will be maybe 1 more day to get all of my comments resolved

proxy/pkg/zdmproxy/streamidmapper.go
```diff
@@ -0,0 +1,107 @@
...
+	streamIdsQueue := make(chan int16, maxStreamIds)
+	for i := int16(0); i < int16(maxStreamIds); i++ {
+		streamIdsQueue <- i
+	}
+	return &internalStreamIdMapper{
+		clusterIds: streamIdsQueue,
+	}
+}
+
+func (csid *internalStreamIdMapper) GetNewIdFor(streamId int16) (int16, error) {
```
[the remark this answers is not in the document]
grighetto: Done

proxy/pkg/zdmproxy/streamidmapper.go
```diff
@@ -0,0 +1,107 @@
...
+	default:
+		return -1, fmt.Errorf("no stream id available")
+	}
+}
+
+func (csid *internalStreamIdMapper) ReleaseId(syntheticId int16) (int16, error) {
+	select {
+	case csid.clusterIds <- syntheticId:
+	default:
+		return syntheticId, fmt.Errorf("stream ids channel full, ignoring id %v", syntheticId)
```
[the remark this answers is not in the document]
grighetto: Fixed

proxy/pkg/zdmproxy/frameprocessor.go
```diff
@@ -0,0 +1,113 @@
...
+	if frame == nil {
+		return frame, nil
+	}
+	var newId, err = sip.mapper.GetNewIdFor(frame.Header.StreamId)
+	if err != nil {
+		return frame, err
+	}
+	if sip.metrics != nil {
+		sip.metrics.Add(1)
+	}
```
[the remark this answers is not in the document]
grighetto: Done

proxy/pkg/zdmproxy/clusterconn.go
grighetto: Fixed

proxy/pkg/zdmproxy/clusterconn.go
grighetto: Done

proxy/pkg/zdmproxy/clusterconn.go
grighetto: Done

proxy/pkg/zdmproxy/streamidmapper.go
grighetto: Fixed

proxy/pkg/zdmproxy/frameprocessor.go
```diff
@@ -0,0 +1,129 @@
...
+
+// releaseStreamId encapsulates the release of the synthetic stream id and the subtraction of the metrics
+// as a reusable function for the different mapper implementation
+func (sip *streamIdProcessor) releaseStreamId(syntheticId int16) (int16, error) {
+	var originalId, err = sip.mapper.ReleaseId(syntheticId)
+	if sip.metrics != nil && err != nil {
+		sip.metrics.Subtract(1)
+	}
+	return originalId, err
+}
```
joao-r-reis, line 129: Move these 2 methods closer to the other processor methods. Currently there are 2 global functions in the middle of them.

joao-r-reis (commented):
Looking good, just waiting for the fix to the protocol error issue before approving the PR.

Also there's still these 2 unresolved comments:
https://github.com/datastax/zdm-proxy/pull/78#discussion_r1040879555
https://github.com/datastax/zdm-proxy/pull/78#discussion_r1034669035

joao-r-reis:
I know that you're looking into the test failure atm but even when you get that test passing we still need a test that verifies that the metric is being added/subtracted. The current test just tests that the number of available stream ids stays at the "max" after executing some requests.

To test this you can create a simulacron test with a primed query that takes like 2 or 3 seconds to give you enough time to assert that the number of available stream ids was subtracted while the request is pending and then it goes back to expected default after the request is done. You can leave this for a follow up PR after we deliver this feature in an early alpha build but that follow up PR should be merged before the next GA so we get proper coverage on this.

integration-tests/metrics_test.go
```diff
@@ -424,9 +423,13 @@ func checkMetrics(
 			require.NotContains(t, lines, fmt.Sprintf("%v", getPrometheusNameWithSuffix(prefix, metrics.OriginRequestDuration, "count")))
 		}
 
-		requireEventuallyContainsLine(t, lines, fmt.Sprintf("%v %v", getPrometheusName(prefix, metrics.AvailableStreamIdsOrigin), streamIdsOrigin))
-		requireEventuallyContainsLine(t, lines, fmt.Sprintf("%v %v", getPrometheusName(prefix, metrics.AvailableStreamIdsTarget), streamIdsTarget))
-		requireEventuallyContainsLine(t, lines, fmt.Sprintf("%v %v", getPrometheusName(prefix, metrics.AvailableStreamIdsAsync), streamIdsAsync))
+		if (successOrigin + successBoth) == 0 {
```
joao-r-reis, line 426: why this if?

proxy/pkg/zdmproxy/clusterconn.go
```diff
@@ -149,13 +148,10 @@ func NewClusterConnector(
 	lastHeartbeatTime := &atomic.Value{}
 	lastHeartbeatTime.Store(time.Now())
 
-	streamIdMetrics, err := GetStreamIdsMetricsByClusterConnector(proxyMetrics, connectorType)
+	streamIdMetrics, err := GetNodeMetricsByClusterConnector(nodeMetrics, connectorType)
```
joao-r-reis, line 151: you don't need this call any more right? you can just use the `nodeMetrics` parameter

proxy/pkg/zdmproxy/frameprocessor.go
grighetto: Done

proxy/pkg/zdmproxy/clusterconn.go
```diff
@@ -236,6 +261,10 @@ func (cc *ClusterConnector) runResponseListeningLoop() {
 		protocolErrOccurred := false
 		for {
 			response, err := readRawFrame(bufferedReader, connectionAddr, cc.clusterConnContext)
+			response, releaseErr := cc.frameProcessor.ReleaseId(response)
+			if releaseErr != nil {
+				log.Error(releaseErr)
```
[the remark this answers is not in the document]
grighetto: João, I implemented our latest decision on this, please have a look.

integration-tests/metrics_test.go
```diff
@@ -41,6 +41,9 @@ var nodeMetrics = []metrics.Metric{
 
 	metrics.OpenOriginConnections,
 	metrics.OpenTargetConnections,
+
+	metrics.OriginUsedStreamIds,
+	metrics.TargetUsedStreamIds,
```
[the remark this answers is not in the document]
grighetto: The `gatherMetrics` checks fail if I add the async metric here. None of the other async metrics have been added either. We can probably improve `gatherMetrics` to make it cope with async, I just didn't want to increase the scope of the changes.

integration-tests/metrics_test.go
grighetto: I thought these would indicate successful request execution, but it looks like they're that and not needed. Removed.

proxy/pkg/zdmproxy/clusterconn.go
grighetto: `nodeMetrics` is the parent struct, it contains `OriginMetrics`, `TargetMetrics` and `AsyncMetrics`. The helper function, already used in other places, will return the proper inner struct depending on the connector type.

integration-tests/metrics_test.go
```diff
@@ -173,9 +176,9 @@ func testMetrics(t *testing.T, metricsHandler *httpzdmproxy.HandlerWithFallback)
 			// 3 on async: AUTH_RESPONSE, STARTUP AND QUERY SELECT
 			// ONLY QUERY is tracked
 			if conf.ReadMode == config.ReadModeDualAsyncOnSecondary {
-				time.Sleep(1000 * time.Millisecond)
+				time.Sleep(10000 * time.Millisecond)
```
joao-r-reis, line 179: 10000 is a bit too much... What failure were you seeing with 1000 before?

proxy/pkg/zdmproxy/clienthandler.go
```diff
@@ -167,13 +167,25 @@ func NewClientHandler(
 	clientHandlerShutdownRequestContext, clientHandlerShutdownRequestCancelFn := context.WithCancel(globalShutdownRequestCtx)
 	requestsDoneCtx, requestsDoneCancelFn := context.WithCancel(context.Background())
 
+	// Initialize stream id processors to manage the ids sent to the clusters
+	originFrameProcessor := newFrameProcessor(conf, nodeMetrics, ClusterConnectorTypeOrigin)
+	targetFrameProcessor := newFrameProcessor(conf, nodeMetrics, ClusterConnectorTypeOrigin)
```
joao-r-reis, line 172: `ClusterConnectorTypeOrigin`

integration-tests/metrics_test.go
grighetto: That must have been a leftover from some experiments I did, but I ran successfully with the original value now on local. Just reverted that.

proxy/pkg/zdmproxy/clienthandler.go
grighetto: Fixed

proxy/pkg/zdmproxy/clientconn.go
```diff
@@ -238,12 +246,12 @@ func checkProtocolError(f *frame.RawFrame, connErr error, protocolErrorOccurred
...
 		if err != nil {
-			return nil, fmt.Errorf("could not generate protocol error response raw frame (%v): %v", protocolErrMsg, err)
+			return nil, fmt.Errorf("could not generate protocol error response raw frame (%v): %v", protocolErrMsg, err), errorCode
 		} else {
-			return rawProtocolErrResponse, nil
+			return rawProtocolErrResponse, nil, errorCode
 		}
 	} else {
-		return nil, connErr
+		return nil, connErr, errorCode
```
joao-r-reis, line 254: `return nil, connErr, 0` here because `errorCode` is being set in the first if even if `protocolErrMsg` is `nil`

proxy/pkg/zdmproxy/clusterconn.go
```diff
@@ -252,6 +263,25 @@ func (cc *ClusterConnector) runResponseListeningLoop() {
 				}
 			}
 
+			// when there's a protocol error, we cannot rely on the returned stream id, the only exception is
+			// when it's a UnsupportedVersion error, which means the Frame was properly parsed but the proxy doesn't
+			// support the protocol version and in that case we can proceed with releasing the stream id in the mapper
+			if response != nil && response.Header.StreamId >= 0 && (err != nil || errCode == ProtocolErrorUnsupportedVersion) {
```
joao-r-reis, line 269: I'm trying to figure out why the `err != nil` here, can you clarify?

grighetto: It was supposed to be `err == nil` as we sync'ed offline (I understand you're changing that yourself with another commit you're doing here. Thanks).

joao-r-reis (approved):

PR #81, opened 2022-12-13
adding ZDM_*_CONTACT_POINTS to configmap

mieslep:


k8s_deploy/configmap.yaml
```diff
@@ -6,9 +6,11 @@ metadata:
 data:
   ZDM_ORIGIN_USERNAME: cassandra
   ZDM_ORIGIN_PASSWORD: cassandra
+  ZDM_ORIGIN_CONTACT_POINTS: "node1.example.internal,node2.example.internal,node3.example.internal"
   ZDM_ORIGIN_PORT: "9042"
   ZDM_TARGET_USERNAME: cassandra
   ZDM_TARGET_PASSWORD: cassandra
+  ZDM_TARGET_CONTACT_POINTS: "newnode1.example.internal,newnode2.example.internal,newnode3.example.internal"
```
weideng1, line 13: ZDM won't be happy if both contact point and scb path is defined. We will need some way to handle that.

PR #82, opened 2022-12-14
K8s helm

weideng1:


k8s_helm_charts/zdm/values.yaml
```diff
@@ -0,0 +1,44 @@
...
+      cpu: 2000m
+      memory: 512Mi
+    requests:
+      cpu: 1000m
+      memory: 512Mi
+  image:
+    repository: datastax/zdm-proxy
+    pullPolicy: IfNotPresent
+    # Overrides the image tag whose default is the chart appVersion.
+    tag: ""
```
jimdickinson, line 25: empty string tag doesn't make a lot of sense for the default, I think? you really *do not* want `latest` for anything more than a short lived debugging container, in k8s

k8s_helm_charts/zdm/templates/secrets.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
```
jimdickinson, line 1: I would not advise the secret lifecycle being tied to the helm release lifecycle. Better to get it out of the helm chart, or at least make it optional here

k8s_helm_charts/zdm/values.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+cdm:
+  resources:
+    limits:
+      cpu: 2000m
+      memory: 4096Mi
+    requests:
+      cpu: 1000m
+      memory: 2048Mi
```
jimdickinson, line 34: good default would have request memory equal limit memory, I think

k8s_helm_charts/README.md
```diff
@@ -0,0 +1,27 @@
+Usage:
+1. Create dedicated namespace for zdmproxy.
+
+    ```kubectl create ns zdmproxy```
+
+2. Import Secure Connect Bundle into a k8s secret called zdmproxy-scb.
+
+    ```kubectl -n zdmproxy create secret generic zdmproxy-scb --from-file=/tmp/secure-connect-origin.zip --from-file=/tmp/secure-connect-target.zip```
```
jimdickinson, line 8: Would recommend getting the below
```
--set origin_username="$prod_astra_user" --set origin_password="$prod_astra_escaped_password" --set target_username="$prod_astra_user" --set target_password="$prod_astra_escaped_password"
```
stuff managed like the SCB.

k8s_helm_charts/zdm/templates/cdm.yaml
```diff
@@ -0,0 +1,64 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: cdm
+  name: cdm-1
```
jimdickinson, line 6: why -1?

k8s_helm_charts/zdm/templates/configmap.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: {{ include "zdm.fullname" . }}
+  namespace: {{ .Values.namespace }}
+data:
+  ZDM_ORIGIN_PORT: "9042"
+  ZDM_TARGET_PORT: "9042"
+  ZDM_PRIMARY_CLUSTER: ORIGIN
+  ZDM_READ_MODE: PRIMARY_ONLY
+  ZDM_LOG_LEVEL: INFO
+  ZDM_PROXY_MAX_CLIENT_CONNECTIONS: "1000"
+  ZDM_METRICS_ENABLED: "true"
```
jimdickinson, line 13: should some of this be helm values? I thought ZDM had a couple different modes and you wouldn't it clobbered on an upgrade?

k8s_helm_charts/zdm/values.yaml
weideng1: So in deployment we have the following reference for the proxy image:
`    spec:
      containers:
      - image: "{{ $.Values.proxy.image.repository }}:{{ $.Values.proxy.image.tag | default $.Chart.AppVersion }}"`

k8s_helm_charts/zdm/templates/configmap.yaml
weideng1: Yeah that's a good reminder. I definitely should address this.

k8s_helm_charts/zdm/templates/secrets.yaml
weideng1: good point, also saw your earlier comment in the README.md. I'll adjust them.

k8s_helm_charts/zdm/values.yaml
weideng1: ok, I'll change them to 2 CPU and 4GB RAM.

weideng1: Should be addressed in the latest commit.

k8s_helm_charts/zdm/templates/secrets.yaml
weideng1: Should be addressed in the latest commit.

k8s_helm_charts/zdm/templates/configmap.yaml
weideng1: Should be addressed in the latest commit.

k8s_helm_charts/zdm/templates/cdm.yaml
weideng1: Should be addressed in the latest commit.

k8s_helm_charts/README.md
weideng1: Should be addressed in the latest commit.

k8s_helm_charts/zdm/values.yaml
```diff
@@ -0,0 +1,46 @@
...
+  primaryCluster: "ORIGIN"
+  maxClientConnections: "1000"
+  metricsEnabled: "true"
+  resources:
+    limits:
+      cpu: 4000m
+      memory: 8192Mi
+    requests:
+      cpu: 4000m
+      memory: 8192Mi
```
joao-r-reis, line 22: 8GB ram seems a bit too agressive for the proxy instances. We haven't benchmarked the memory usage of the proxy in a long time but the last time we did we didn't see usage go over 500MB. 1GB should be fine but we could also go for 2GB since we don't have recent data so it's better to be extra safe.

k8s_helm_charts/zdm/templates/deployment.yaml
```diff
@@ -0,0 +1,110 @@
...
+        - name: ZDM_METRICS_ADDRESS
+          valueFrom:
+            fieldRef:
+              fieldPath: status.podIP
+        - name: ZDM_METRICS_PORT
+          value: "14001"
+        - name: ZDM_PROXY_TOPOLOGY_INDEX
+          value: {{ $index | quote }}
+        - name: ZDM_PROXY_TOPOLOGY_ADDRESSES
+          value: {{ $service_addresses }}
```
grighetto, line 63: Are `service_addresses` actual IP address or DNS names?
I remember I discussed this with @joao-r-reis some time ago and since these are used to spoof the peers addresses in the `system.peers` table and the data type is `inet`, it could cause problems to have it as DNS names because of the incompatible type.
At least that was the case back in February when I added the ZDM proxy in a Helm chart, I'm not sure if things changed since then.

weideng1: Those environmental variables are exposed as IP addresses. They're a built-in support from k8s and all pods have access to these environmental variables to get to the IP addresses of all running services. See this article for more details: https://learn.microsoft.com/en-us/visualstudio/bridge/kubernetes-environment-variables

grighetto: Ok, cool. Yeah, I'm familiar with that, I just didn't notice the env var **value** was being extracted with the `$(...)` notation, I thought it was using the var name as the output. That looks good then 👍

k8s_helm_charts/zdm/values.yaml
weideng1: changed to 2GB

k8s_helm_charts/zdm/values.yaml
jimdickinson: 👍

jimdickinson (approved):

PR #86, opened 2022-12-22
Pin go version to 1.19

joao-r-reis:
Some places are setting go version to 1.19, others are setting the version to 1.19.2 and others aren't setting it at all. Also go.mod specifies 1.16 which is not compatible with the recent change to import a sys library (https://github.com/golang/go/issues/51706).


grighetto (approved):
I checked and the tests are already pinned to 1.19, so this looks good to me 👍

PR #89, opened 2022-12-28
New version of zdm-proxy helm chart using StatefulSets

weideng1:
@jimdickinson @bradfordcp when you're back from the holidays, could you please help to take a look at the helm chart in this PR? Compared to the original version, the main improvement in this PR is that I've now started to use STS to manage all proxy pods, while keeping the services to route traffic to individual proxy pods via their names. This enables regular rolling restart using `kubectl -n zdmproxy rollout restart sts/zdm-proxy` with the appropriate wait time in between. Changes to values.yaml or `--set` values will trigger all proxy pods to roll-restart, so we can easily scale out and scale in.

I also removed configmap.yaml as all of the values in there can be directly retrieved from helm chart's values.yaml. This way, changes to these values can be directly detected by StatefulSets and automatically trigger rolling restart of all proxy pods without causing any downtime.


lorantfecske-bud:
I was going through the PR as we are planing to utilize the helm chart to deploy the zdm-proxy to our clusters.

I have few questions about this PR, it seems to me that you have added a new stateful set and set the services to point to the pods in the stateful set. However the deployment is still there, so if I understand this correctly, alongside the new statefulset the zdm proxy is also going to be deployed as a deployment, but no service will point to the deployment(s) anymore. 

I'm also not 100% clear why the CDM deployment has been turned into a simple pod. We are using preemptive instances and they are recycled frequently. Based on kubernetes docs if the node is going away the pod is gone. 
`The Pod remains on that node until the Pod finishes execution, the Pod object is deleted, the Pod is evicted for lack of resources, or the node fails.` Source: https://kubernetes.io/docs/concepts/workloads/pods/

weideng1:
> I have few questions about this PR, it seems to me that you have added a new stateful set and set the services to point to the pods in the stateful set. However the deployment is still there, so if I understand this correctly, alongside the new statefulset the zdm proxy is also going to be deployed as a deployment, but no service will point to the deployment(s) anymore.

@lorantfecske-bud The latest commit in this branch `k8s-helm` has removed deployment completely, so zdm-proxy will use statefulsets, instead of deployment.

weideng1:
> I'm also not 100% clear why the CDM deployment has been turned into a simple pod.

@lorantfecske-bud We're planning to take CDM (Cassandra Data Migrator) out of the zdm-proxy helm chart, as it is a completely separate open source project https://github.com/datastax/cassandra-data-migrator and needs to have its own deployment and lifecycle management approach. For now, you can use the following command to run CDM in your k8s cluster without the helm chart:

```
kubectl run cdm --image=datastax/cassandra-data-migrator:latest
```

k8s_helm_charts/zdm/templates/sts.yaml
```diff
@@ -0,0 +1,134 @@
...
+          initialDelaySeconds: 10
+          periodSeconds: 10
+        ports:
+          - name: httpproxy
+            containerPort: 14001
+            protocol: TCP
+          - name: cql
+            containerPort: 9042
+            protocol: TCP
+        resources:
```
jsanda, line 60: I would just specify the resources as a yaml block:

```yaml
resources:
  {{- toYaml .Values.proxy.resources | nindent 12 }}  # actual value for nindent may vary
```

k8s_helm_charts/zdm/templates/sts.yaml
```diff
@@ -0,0 +1,134 @@
...
+{{- $zdm_labels := include "zdm.labels" . -}}
+{{- $zdm_selectorLabels := include "zdm.selectorLabels" . -}}
+
+# calculate a variable that contains all proxy service addresses
+{{ $service_addresses := "" -}}
+{{- range $index := until (.Values.proxy.count | int) -}}
+    {{- $service_addresses = printf "%s,$(%s_%s_SERVICE_HOST)" $service_addresses ($zdm_fullname | upper | replace "-" "_") ($index | toString) -}}
+{{- end -}}
+{{- $service_addresses = $service_addresses | trimPrefix "," -}}
+
```
jsanda, line 12: While it isn't specific to this PR, I recommend that you document every property in values.yaml. Think of as your user-facing, client API :)

k8s_helm_charts/zdm/templates/service.yaml
```diff
@@ -21,6 +21,6 @@ spec:
       name: cql
   selector:
     {{- $zdm_selectorLabels | nindent 4 }}
-    app: {{ $zdm_fullname }}-{{ $index }}
+    statefulset.kubernetes.io/pod-name: {{ $zdm_fullname }}-{{ $index }}
```
jsanda, line 24: Why are you trying to match a per-pod label?

lorantfecske-bud:
@weideng1 thanks for then answers

k8s_helm_charts/zdm/templates/sts.yaml
weideng1: Good recommendation. I've now completely revised values.yaml based on the best practices described in [this helm documentation page](https://helm.sh/docs/chart_best_practices/values/).

k8s_helm_charts/zdm/templates/service.yaml
weideng1: @jsanda The reason for us to create individual services and match them 1:1 at pod level is because unlike C*, zdm-proxy processes don't have gossip protocol to detect/notify topology changes among themselves; whatever set of IP addresses and their respective ordinal index (see [here](https://github.com/datastax/zdm-proxy/blob/f893984498547d5181c0216e955e01949ff5b88b/proxy/pkg/config/config.go#L27-L28)) specified at the start of the proxy process is going to stuck for the rest of its lifecycle. To allow dynamical IP address change to the pods due to reschedule/cordon/crash, we decided to for N number of proxies, implement N number of services, and them map 1-on-1. Given that pods managed by k8s Deployment object doesn't have a static pod name, we switched to using StatefulSets. In StatefulSets, each pod has a unique/static pod-name that we can map the service to. This will allow orderly rolling restart of the zdm-proxy pods to happen.

k8s_helm_charts/zdm/templates/sts.yaml
```diff
@@ -5,46 +5,71 @@
...
     spec:
       containers:
-      - image: "{{ $.Values.proxy.image.repository }}:{{ $.Values.proxy.image.tag | default $.Chart.AppVersion }}"
-        name: {{ $zdm_fullname }}-{{ $index }}
+      - image: "{{ $.Values.imageRepository }}:{{ $.Values.imageTag | default $.Chart.AppVersion }}"
+        name: {{ $zdm_fullname }}
+        command:
+        - sh
+        - "-c"
+        - "ZDM_PROXY_TOPOLOGY_INDEX=`echo ${HOSTNAME##*-}` /main"
```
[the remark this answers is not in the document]
bradfordcp: Hmm, it looks like you're [extracting the ordinal index which is not available to the pod directly](https://stackoverflow.com/a/50751423/267475). It's frustrating that this isn't available. (I see the link to https://github.com/kubernetes/kubernetes/issues/40651 as well)

bradfordcp (commented):
Popping some comments in, but no glaring blockers from me. I expect @jsanda will handle proper approval.

k8s_helm_charts/zdm/templates/sts.yaml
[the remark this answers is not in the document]
weideng1: @bradfordcp Yes exactly. That was the route I explored and ended up using the workaround as you pointed out. Should we resolve this comment?

PR #99, opened 2023-01-20
ZDM-522 Add size limit to PS Cache

### LINKED ISSUE #98: Add a size limit to the proxy's prepared statement cache
The proxy's PS "cache" is implemented as a map that holds known preparedIDs and a boolean indicating whether a prepared statement is a read or a write. When prepared statements are used appropriately by client applications, this should not pose any issue in terms of map size or churn.

However, it is not uncommon for client applications to misuse prepared statements. While re-preparing the same statement over and over would be harmless, a client application that does something like hardcoding parameters in the statement for every call and then prepare it would cause a different preparedId every time. This would make the map grow very large, resulting in performance degradation and high memory usage and, in the worst cases, making the the proxy crash.

To prevent this, we need to:
- Add a test to reproduce the issue
- Add a limit to the size of the map (this can be quite high, as the footprint of each element is very small, but at least it is not unbounded)
- Change the PS cache maps to use an LRU cache, such as the one provided by the Hashicorp LRU library.

alicel:
- Replaced all three maps in the PSCache struct with the LRU cache provided by Hashicorp.
- Added a configuration variable that sets a default size limit of 5000 for now (the footprint of each element is small so it could be much higher, but applications should not need to create so many prepared statements anyway)
- Added a unit test that verifies that the PS Cache behaves as expected


proxy/pkg/zdmproxy/pscache.go
```diff
@@ -64,31 +81,37 @@ func (psc *PreparedStatementCache) StoreIntercepted(preparedResult *message.Prep
 func (psc *PreparedStatementCache) Get(originPreparedId []byte) (PreparedData, bool) {
 	psc.lock.RLock()
 	defer psc.lock.RUnlock()
```
joao-r-reis, line 83: We need to reevaluate these locks. This new cache implementation already does locking so it would be great if we could remove our locks.

proxy/pkg/zdmproxy/pscache.go
```diff
@@ -4,33 +4,50 @@ import (
...
-	}
+	}, nil
 }
 
 func (psc PreparedStatementCache) GetPreparedStatementCacheSize() float64 {
 	psc.lock.RLock()
 	defer psc.lock.RUnlock()
 
-	return float64(len(psc.cache) + len(psc.interceptedCache))
+	//return float64(len(psc.cache) + len(psc.interceptedCache))
```
joao-r-reis, line 49: delete

proxy/pkg/zdmproxy/pscache.go
```diff
@@ -42,8 +59,8 @@ func (psc *PreparedStatementCache) Store(
 	psc.lock.Lock()
 	defer psc.lock.Unlock()
 
-	psc.cache[originPrepareIdStr] = NewPreparedData(originPreparedResult, targetPreparedResult, prepareRequestInfo)
-	psc.index[targetPrepareIdStr] = originPrepareIdStr
+	psc.cache.Add(originPrepareIdStr, NewPreparedData(originPreparedResult, targetPreparedResult, prepareRequestInfo))
+	psc.index.Add(targetPrepareIdStr, originPrepareIdStr)
```
joao-r-reis, line 63: Hmm what happens if the limit is reached and the key that is selected to be evicted on one cache is different from the key that was selected on the other cache?

I'm not sure if bad behavior can happen if these two caches are out of sync or if it's fine...

If we do need both caches to be in sync then a potential alternative would be to provide an eviction callback to the `cache` that removes that key from the `index` cache. This way we could revert the `index` cache to a normal `map` (with our own locks) and perform the eviction ourselves.

joao-r-reis (commented):
Added a couple of comments.

integration-tests/setup/testcluster.go
```diff
@@ -437,6 +437,7 @@ func NewTestConfig(originHost string, targetHost string) *config.Config {
 
 	conf.ProxyMaxClientConnections = 1000
 	conf.ProxyMaxStreamIds = 2048
+	conf.ProxyMaxPreparedStatementCacheSize = 5000
```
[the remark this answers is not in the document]
alicel: The limit is based on size, rather than number of prepared statements: `The default calculated value is 1/256th of the heap or 10 MB, whichever is greater`.

When I set the default to 5000, my intention was to keep the PS cache memory footprint relatively small, given that the proxy usually runs on instances with limited resources. It is quite unlikely for applications to create a large number of prepared statements if they are using them correctly, so even if a user has multiple applications using the proxy I thought that this should be a reasonable value. On the other hand, the footprint of each statement in the proxy PS cache maps is small, so choosing a default value that is a bit higher should also be fine.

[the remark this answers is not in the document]
joao-r-reis: So for a 16GB heap the limit would be about 60MB, I think it would be good to get an estimate of what size an average prepared statement has on our cache so we can come up with a good value for this limit instead of guessing blindly. 5000 sounds a bit too low but without any data on the size that each statement takes I'm just guessing blindly. Is there server metrics for the prepared cache size? We could use those metrics in a benchmark to get some data that would help us come up with a good limit.

Issue #100, opened 2023-02-07
Allow other password authenticator to work


weideng1:
The [current code](https://github.com/datastax/zdm-proxy/blob/f893984498547d5181c0216e955e01949ff5b88b/proxy/pkg/zdmproxy/auth.go#L98-L106) only accepts `com.datastax.bdp.cassandra.auth.DseAuthenticator` and `org.apache.cassandra.auth.PasswordAuthenticator`, which unnecessarily limits the kind of CQL data stores that can be used as `Origin` or `Target`. The proposed change takes [the same approach that CQL python driver takes](https://github.com/datastax/python-driver/blob/master/cassandra/auth.py#L259-L263), i.e. if it's not a DSEAuthenticator, it will assume the CQL data store uses a compatible password authenticator. When we need to handle more non-password authenticators, this segment will likely change back to switch statement, but that's going to be a bigger change any way.

┆Issue is synchronized with this [Jira Task](https://datastax.jira.com/browse/ZDM-530) by [Unito](https://www.unito.io)
┆Components: Proxy
┆Issue Number: ZDM-530

ims-swilkinson:
We've hit this problem trying to migrate from DSE to Astra. For Astra it needs to support the "org.apache.cassandra.auth.AstraAuthenticator" authenticator. This seems to be compatible with "org.apache.cassandra.auth.PasswordAuthenticator" which is already supported.

joao-r-reis:
> We've hit this problem trying to migrate from DSE to Astra. For Astra it needs to support the "org.apache.cassandra.auth.AstraAuthenticator" authenticator. This seems to be compatible with "org.apache.cassandra.auth.PasswordAuthenticator" which is already supported.

the PR hasn't been merged yet but the fix is available in docker image `datastax/zdm-proxy:sha-33db4fd` if needed

PR #101, opened 2023-02-07
#100 allow other password authenticator to work ZDM-530

weideng1:


joao-r-reis:
Did you check what each driver is doing? I checked C# and it ignores the name like the proxy would with this change. I assume python is the same and java driver 3.x too due to the tests that you ran but what about java driver 4.x?

joao-r-reis:
@absurdfarce raised a valid point that we may not want to assume things about the authenticator so we could add an additional `if` here for the particular authenticator that we're trying to support here. Authenticators at the driver level are chosen by the user but this one in zdm proxy is pretty much internal because we don't really support any configuration on it.

In the end it may just be better to do like the drivers do and assume that if it's not the dse authenticator then the plaintext one should be used. @absurdfarce what's your opinion on this?

joao-r-reis:
Also please create a GH issue and add this item to the changelog

weideng1:
> Also please create a GH issue and add this item to the changelog

@joao-r-reis the GH issue was created as #100 before I opened this PR.

absurdfarce:
Mentioned in Slack, mentioning here for sake of completeness.  We've looked at the behaviour in the Python driver as well as 3.x and 4.x of the Java driver; all three follow the same basic pattern outlined by @weideng1 here.

absurdfarce:
So the one major difference between the examples we've looked at so far is that (at least for the Python, Java 3.x and Java 4.x cases) each driver implements the authentication logic as a pluggable component.  What I've described then is the _default_ behaviour for the unmodified drivers but consumers are always free to define their own authn logic and implement a custom AuthProvider (or equivalent) to make it happen.

What we've been discussing on the zdm side isn't as pluggable.  So there is always the _potential_ for somebody to come along with a custom authenticator we can't support via the logic described above.

I don't see any obvious way around that besides supporting a pluggable authentication model in zdm as well.  But that's a different question from whether we should make the change outlined here.  This model changes us to "use the DSE authenticator if specified otherwise assume you have something that supports plaintext auth".  As a default case that doesn't seem unreasonable to me.

joao-r-reis:
> > Also please create a GH issue and add this item to the changelog
> 
> @joao-r-reis the GH issue was created as #100 before I opened this PR.

Add it to the changelog on this PR please

joao-r-reis:
And add `Fixes #100` to the PR description so it links and closes the GH issue automatically

joao-r-reis:
And we should create a new milestone for an upcoming version that will have this change, if we are not sure what the version will be yet it can be created as TBD or 2.next or 2.x (and we will rename it when we decide which version it will be)

jeromatron:
So the remaining item is just to add it to the changelog?  After that, can we merge this and it can await a next release?

joao-r-reis (approved):
When CI is done, feel free to merge

PR #114, opened 2024-06-06
Update NoSQLBench version to leverage the latest 5.21.1

msmygit:
- Updates minor documentation items.
- Leverages NoSQLBench to the latest `5.21.x` version.
- Adds dependabot that will help keep the GHA runner versions up to date so as to avoid warnings like as below,
```
Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v2, test-summary/action@v1. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.
```

### DIFF
--- a/CHANGELOG/CHANGELOG-2.1.md
+++ b/CHANGELOG/CHANGELOG-2.1.md
@@ -22,3 +22,5 @@ When cutting a new release, update the `unreleased` heading to the tag being gen
 * [#69](https://github.com/datastax/zdm-proxy/issues/69) Client connection can be closed before proxy returns protocol error
 * [#76](https://github.com/datastax/zdm-proxy/issues/76) Log error when closing connection
 * [#74](https://github.com/datastax/zdm-proxy/issues/74) Handshakes with auth enabled can deadlock if multiple handshakes are happening concurrently
+
+---
--- a/compose/nosqlbench-entrypoint.sh
+++ b/compose/nosqlbench-entrypoint.sh
@@ -56,22 +56,22 @@ java -jar /nb5.jar \
   errors=counter,retry \
   -v
 
-echo "Running NoSQLBench VERIFY job on ORIGIN"
-java -jar /nb5.jar \
-  --show-stacktraces \
-  --report-csv-to /source/verify-origin \
-  /source/nb-tests/cql_nb_activity.yaml \
-  verify \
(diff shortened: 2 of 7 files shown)


CHANGELOG/CHANGELOG-2.1.md
```diff
@@ -4,7 +4,7 @@ Changelog for the ZDM Proxy, new PRs should update the `unreleased` section.
 
 When cutting a new release, update the `unreleased` heading to the tag being generated and date, like `## vX.Y.Z - YYYY-MM-DD` and create a new placeholder section for  `unreleased` entries.
 
-## Unreleased
+## Released
```
joao-r-reis, line 7: This was correct. We use the `Unreleased` section when we have things merged in `main` that have not been released yet.

msmygit: now since it was released, https://github.com/datastax/zdm-proxy/releases/tag/v2.1.0 shouldn't we move it to a released section?

joao-r-reis: there is no "Released" section. There is one section per release and one section for everything Unreleased (no version exists)

compose/nosqlbench-entrypoint.sh
```diff
@@ -17,62 +21,57 @@ test_conn zdm_tests_origin
...
-java -jar /nb.jar \
-  --show-stacktraces \
-  --report-csv-to /source/verify-origin \
-  /source/nb-tests/cql-nb-activity.yaml \
-  verify \
-  driver=cqld3 \
-  hosts=zdm_tests_origin \
-  localdc=datacenter1 \
-  -v
+#echo "Running NoSQLBench VERIFY job on ORIGIN"
```
grighetto, line 59: Without the VERIFY step, this test is pretty much useless because this is what actually compares the stored data against the pseudo-random generated data.
I don't have the context why the NB is happening in this PR (if it's for some new feature or compatibility with newer C* versions), but if we could still run the old NB with the verify step in addition to this, until it's support in the latest NB version, I think it would be ideal.

msmygit: Understood. NB 5.21 doesnt support `verify` feature yet.

Issue #91, opened 2023-01-09
Use the control connection address for the control host in the cluster topology metadata seen by the proxy


alicel:
When the proxy connects to Origin and Target, it builds a view of the metadata of each using the data from their system.local and system.peers tables.

These queries are run on the control host after the control connection is established. While the RPC address of the peers is always available in the system.peers table, early C* 2.1 versions did not contain the RPC address of the local node in system.local. 

At the moment we rely on the RPC address from system.local to populate the address of the control host when we create the topology information in the proxy. This means that, in case of clusters on C* versions older than 2.1.6, the address of the control host in the proxy’s topology view is going to be null. Apart from being generally incorrect, this causes a practical issue if a proxy instance selects its control host as one of its assigned hosts, as it will be unable to connect to this malformed assigned host.

To fix this for all versions, we should be using the control host address (which is always known, as we are connecting to it in the first place) when populating the control host metadata in the cluster topology, instead of relying on the address in system.local. This is in line with what the Java driver does.

For this change to work efficiently, we must ensure that we always use `hostID` as the node identifier and consider the IP address just like a regular piece of metadata about the node.

joao-r-reis:
Also worth noting that this is useful for deployments that have specific network conditions where the contact point address does not match the one on `system.local` although this is probably only useful for single node deployments because it would result in the proxy being able to connect to the local host only.

PR #119, opened 2024-06-17
120: Support protocol v2

### LINKED ISSUE #120: Support protocol version 2
Currently the control connection only connects using v3 and max stream ids defaults to 2048 which is much higher than the v2 limit (128).

Both of these 2 issues need to be addressed to properly support v2.

┆Issue is synchronized with this [Jira Task](https://datastax.jira.com/browse/ZDM-610) by [Unito](https://www.unito.io)
┆Components: Proxy
┆Issue Number: ZDM-610

lukasz-antoniak:
fix #120


proxy/pkg/zdmproxy/controlconn.go
```diff
@@ -372,6 +368,52 @@ func (cc *ControlConn) openInternal(endpoints []Endpoint, ctx context.Context) (
...
+	for {
+		tcpConn, _, err := openConnection(cc.connConfig, endpoint, ctx, false)
+		if err != nil {
+			log.Warnf("Failed to open control connection to %v using endpoint %v: %v",
+				cc.connConfig.GetClusterType(), endpoint.GetEndpointIdentifier(), err)
+			return nil, err
+		}
+		newConn := NewCqlConnection(tcpConn, cc.username, cc.password, ccReadTimeout, ccWriteTimeout, cc.conf)
+		err = newConn.InitializeContext(protoVer, ctx)
+		if err != nil && strings.Contains(err.Error(), "Invalid or unsupported protocol version") {
```
joao-r-reis, line 382: Before checking the message we should check if the error is a `Protocol Error`

integration-tests/connect_test.go
```diff
@@ -45,6 +46,40 @@ func TestGoCqlConnect(t *testing.T) {
...
+	testSetup, err := setup.NewSimulacronTestSetupWithConfig(t, c)
+	require.Nil(t, err)
+	defer testSetup.Cleanup()
+
+	// Connect to proxy as a "client"
+	proxy, err := utils.ConnectToClusterUsingVersion("127.0.0.1", "", "", 14002, 3)
+
+	if err != nil {
+		t.Log("Unable to connect to proxy session.")
+		t.Fatal(err)
```
joao-r-reis, line 66: why not `t.Fatal("Unable to connect to proxy session: ", err)` just like you did a few lines below

integration-tests/connect_test.go
```diff
@@ -45,6 +46,40 @@ func TestGoCqlConnect(t *testing.T) {
 	require.Equal(t, "fake", iter.Columns()[0].Name)
 }
 
+func TestProtocolVersionNegotiation(t *testing.T) {
+	testCassandraVersion := env.CassandraVersion
+	env.CassandraVersion = "2.1" // downgrade C* version for protocol negotiation test
```
joao-r-reis, line 51: we shouldn't do this, we should just add a function that creates a simulacron cluster with the provided version instead of using the env var. Maybe this should even have multiple test cases where you provide the inputs: `C* version, expected protocol version in control connection`, and you can create a function in the control connection to return the current protocol version to assert on this test

proxy/pkg/zdmproxy/cqlconn.go
```diff
@@ -440,7 +442,7 @@ func (c *cqlConn) Execute(msg message.Message, ctx context.Context) (message.Mes
 
 func (c *cqlConn) SendHeartbeat(ctx context.Context) error {
 	optionsMsg := &message.Options{}
-	heartBeatFrame := frame.NewFrame(ccProtocolVersion, -1, optionsMsg)
+	heartBeatFrame := frame.NewFrame(c.protocolVersion, -1, optionsMsg)
```
joao-r-reis, line 445: access to `c.protocolVersion` should probably be atomic, maybe using `atomic.Value`

proxy/pkg/zdmproxy/cqlconn.go
```diff
@@ -59,6 +59,7 @@ type cqlConn struct {
 	eventHandlerLock      *sync.Mutex
 	authEnabled           bool
 	frameProcessor        FrameProcessor
+	protocolVersion       primitive.ProtocolVersion
```
joao-r-reis, line 62: this should probably be `atomic.Value`

proxy/pkg/zdmproxy/controlconn.go
```diff
@@ -372,6 +368,52 @@ func (cc *ControlConn) openInternal(endpoints []Endpoint, ctx context.Context) (
...
+			}
+			protoVer = downgradeProtocol(protoVer)
+			log.Infof("Downgrading protocol version: %v", protoVer)
+			if protoVer == 0 {
+				// we cannot downgrade anymore
+				return nil, err
+			}
+			continue // retry lower protocol version
+		} else {
+			cc.protocolVersion = protoVer
```
joao-r-reis, line 397: I don't think we should set this if there was an error.

proxy/pkg/config/config.go
```diff
@@ -21,6 +21,7 @@ type Config struct {
 	ReplaceCqlFunctions     bool   `default:"false" split_words:"true"`
 	AsyncHandshakeTimeoutMs int    `default:"4000" split_words:"true"`
 	LogLevel                string `default:"INFO" split_words:"true"`
+	ProtocolVersion         uint   `default:"3" split_words:"true"`
```
joao-r-reis, line 24: This should be `ControlConnMaxProtocolVersion` to make it clear we are not messing with the protocol version on the non control connections. And this setting is confusing because at the moment it is more of a "MaxProtocolVersion" instead of one that is "forced" which I as a user would expect (I think) if I just read the name "Protocol Version".

proxy/pkg/zdmproxy/controlconn.go
```diff
@@ -54,11 +54,11 @@ type ControlConn struct {
 	protocolEventSubscribers map[ProtocolEventObserver]interface{}
 	authEnabled              *atomic.Value
 	metricsHandler           *metrics.MetricHandler
+	protocolVersion          primitive.ProtocolVersion
```
joao-r-reis, line 57: Do we need this? Can't we just grab the protocol version from the underlying `cqlConn`?

proxy/pkg/zdmproxy/controlconn.go
```diff
@@ -387,7 +429,7 @@ func (cc *ControlConn) Close() {
 }
 
 func (cc *ControlConn) RefreshHosts(conn CqlConnection, ctx context.Context) ([]*Host, error) {
-	localQueryResult, err := conn.Query("SELECT * FROM system.local", GetDefaultGenericTypeCodec(), ccProtocolVersion, ctx)
+	localQueryResult, err := conn.Query("SELECT * FROM system.local", GetDefaultGenericTypeCodec(), cc.protocolVersion, ctx)
```
joao-r-reis, line 432: We don't need to pass a protocol version anymore, the cql connection itself has it.

proxy/pkg/zdmproxy/controlconn.go
```diff
@@ -410,7 +452,7 @@ func (cc *ControlConn) RefreshHosts(conn CqlConnection, ctx context.Context) ([]
 		}
 	}
 
-	peersQuery, err := conn.Query("SELECT * FROM system.peers", GetDefaultGenericTypeCodec(), ccProtocolVersion, ctx)
+	peersQuery, err := conn.Query("SELECT * FROM system.peers", GetDefaultGenericTypeCodec(), cc.protocolVersion, ctx)
```
joao-r-reis, line 455: same as above

integration-tests/connect_test.go
joao-r-reis: Also I'm not sure if simulacron will respect the C* version when considering which protocol versions to use. We might have to use the "fake" test server (don't really need CCM here, the "fake" test server should work just fine and would make the tests much faster).

proxy/pkg/config/config.go
```diff
@@ -16,11 +16,12 @@ type Config struct {
...
-	ReadMode                string `default:"PRIMARY_ONLY" split_words:"true"`
-	ReplaceCqlFunctions     bool   `default:"false" split_words:"true"`
-	AsyncHandshakeTimeoutMs int    `default:"4000" split_words:"true"`
-	LogLevel                string `default:"INFO" split_words:"true"`
+	PrimaryCluster                string `default:"ORIGIN" split_words:"true"`
+	ReadMode                      string `default:"PRIMARY_ONLY" split_words:"true"`
+	ReplaceCqlFunctions           bool   `default:"false" split_words:"true"`
+	AsyncHandshakeTimeoutMs       int    `default:"4000" split_words:"true"`
+	LogLevel                      string `default:"INFO" split_words:"true"`
+	ControlConnMaxProtocolVersion uint   `default:"3" split_words:"true"`
```
joao-r-reis, line 24: I'd probably put this as a string because what if someone wants to set `Dsev1` or `Dsev2`? They would have to provide the actual protocol code for it which users don't know. And this config setting should then be validated and startup should be aborted if the provided protocol version string can't be converted to an actual protocol version (and we should disallow anything below v2 and anything higher than v4 except dsev1 and dsev2)

integration-tests/connect_test.go
```diff
@@ -45,6 +46,34 @@ func TestGoCqlConnect(t *testing.T) {
 	require.Equal(t, "fake", iter.Columns()[0].Name)
 }
 
+func TestProtocolVersionNegotiation(t *testing.T) {
+	c := setup.NewTestConfig("", "")
```
joao-r-reis, line 50: we're still missing the multiple test cases that I requested. Feel free to look at other tests that contain "sub tests". We should do the same here with a matrix of C* versions and protocol versions and expected outcomes.

integration-tests/utils/testutils.go
```diff
@@ -116,9 +116,9 @@ func CheckMetricsEndpointResult(httpAddr string, success bool) error {
 	return nil
 }
 
-// ConnectToCluster is used to connect to source and destination clusters
-func ConnectToCluster(hostname string, username string, password string, port int) (*gocql.Session, error) {
+func ConnectToClusterUsingVersion(hostname string, username string, password string, port int, protoVersion int) (*gocql.Session, error) {
 	cluster := NewCluster(hostname, username, password, port)
+	cluster.ProtoVersion = protoVersion
```
joao-r-reis, line 121: keep in mind this skips protocol version negotiation on gocql.

proxy/pkg/zdmproxy/response.go
```diff
@@ -37,3 +41,20 @@ func (r *Response) GetStreamId() int16 {
 		return r.requestFrame.Header.StreamId
 	}
 }
+
+type ResponseError struct {
```
joao-r-reis, line 45: I'd probably add this type near the code of `cqlconnection` or the control connection because it really only applies there right? I don't think it is used outside of the control connection.

proxy/pkg/zdmproxy/cqlconn.go
```diff
@@ -237,6 +239,8 @@ func (c *cqlConn) InitializeContext(version primitive.ProtocolVersion, ctx conte
 		return fmt.Errorf("failed to perform handshake: %w", err)
 	}
 
+	c.protocolVersion = &atomic.Value{}
```
joao-r-reis, line 242: I think we should set this field in `NewCqlConnection` and set the initial protocol version there as well so there's no confusion as to whether this field is being modified after the connection object is created.

joao-r-reis:
The purpose of the ticket is to support v2, protocol negotiation in the control connection is just a part of it so I suggest modifying the name of the PR. 

There's another part that is missing here which is modifying the maximum number of stream ids and that part might be more tricky and will require you to deep dive a bit more into the other part of the codebase (cluster connector, client connector, client handler). 

Modifying the number of stream ids in the control connection is probably not hard and it's on the part of the code that you're already working with so you can start there. But we also need to set the max number of stream ids based on the protocol version on the non control connections too. At the moment there's a config setting for it that defaults to 2048 but with v2 the connection only supports up to 128 stream ids.

joao-r-reis:
Also we should make sure we have tests covering v2 protocol to validate this PR.

proxy/pkg/zdmproxy/response.go
lukasz-antoniak: I was actually hoping this type to be reusable when somebody needs to raise error and carry response frame. Therefore, I also tried to think of some reusable name.

integration-tests/utils/testutils.go
lukasz-antoniak: Hmm, without this connection to Simulacron fails in `Cluster2.1_MaxCCProtoVer4_NegotiatedProtoVer3`.

joao-r-reis (commented):
Looking good, left some comments.

joao-r-reis:
Also, the stream ids are still left unchanged, we need to override the max stream ids config setting when it's v2 in case it's above what v2 supports. And need a test for it.

go.sum
```diff
@@ -119,12 +162,20 @@ golang.org/x/sys v0.0.0-20181116152217-5ac8a444bdc5/go.mod h1:STP8DvDyc/dI5b8T5h
...
+golang.org/x/sys v0.6.0/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
+golang.org/x/sys v0.12.0/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
+golang.org/x/sys v0.21.0 h1:rF+pYz3DAGSQAxAu1CbC7catZg4ebC4UIeIhKxBZvws=
+golang.org/x/sys v0.21.0/go.mod h1:/VUhepiaJMQUp4+oa/7Zr1D23ma6VTLIYjOOTFZPUcA=
 golang.org/x/text v0.3.0/go.mod h1:NqM8EUOU14njkJ3fqMW+pc6Ldnwhi/IjpwHt7yyuwOQ=
 golang.org/x/tools v0.0.0-20190828213141-aed303cbaa74/go.mod h1:b+2E5dAYhXwXZwtnZ6UAqBI28+e2cm9otk0dWdXHAEo=
 golang.org/x/xerrors v0.0.0-20190717185122-a985d3407aa7/go.mod h1:I/5z698sn9Ka8TeJc9MKroUUfqBBauWjQqLJ2OPfmY0=
 golang.org/x/xerrors v0.0.0-20191204190536-9bdfabe68543/go.mod h1:I/5z698sn9Ka8TeJc9MKroUUfqBBauWjQqLJ2OPfmY0=
+google.golang.org/protobuf v1.26.0-rc.1/go.mod h1:jlhhOSvTdKEhbULTjvd4ARK9grFBp09yW+WbY/TyQbw=
+google.golang.org/protobuf v1.26.0/go.mod h1:9q0QmTI4eRPtz6boOQmLYwt+qCgq0jsYwAQnmE0givc=
```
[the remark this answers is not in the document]
lukasz-antoniak: Sorry missed that. You are 100% correct.

proxy/pkg/zdmproxy/cqlparser_adv_workloads_utils_test.go
```diff
@@ -49,8 +51,8 @@ func buildQueryMessageForTests(queryString string) *message.Query {
 			PageSize:          5000,
 			PageSizeInBytes:   false,
 			PagingState:       nil,
-			SerialConsistency: &primitive.NillableConsistencyLevel{Value: primitive.ConsistencyLevelLocalSerial},
-			DefaultTimestamp:  &primitive.NillableInt64{Value: 1647023221311969},
+			SerialConsistency: &serialConsistency,
+			DefaultTimestamp:  &defaultTimestamp,
```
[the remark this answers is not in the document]
lukasz-antoniak: `go-cassandra-native-protocol` changed the type to pointer.

integration-tests/utils/testutils.go
lukasz-antoniak: I did not test it end-to-end, but looking at GoCQL code, it seems it does not support protocol version negotiation. See https://github.com/gocql/gocql/blob/master/control.go#L219. It tries always V4. Error raised with Simulacron test is: `gocql: unable to create session: unable to discover protocol version: unexpected protocol version in response: got 3 expected 4`. Simulacron returns V3, but driver does not automatically downgrade.

joao-r-reis: gocql absolutely supports protocol version negotiation but you're skipping it by telling gocql to use a specific version, you should remove that from that test

integration-tests/protocolv2_test.go
```diff
@@ -0,0 +1,171 @@
...
+	"github.com/datastax/go-cassandra-native-protocol/frame"
+	"github.com/datastax/go-cassandra-native-protocol/message"
+	"github.com/datastax/go-cassandra-native-protocol/primitive"
+	"github.com/datastax/zdm-proxy/integration-tests/setup"
+	"github.com/stretchr/testify/require"
+	"net"
+	"testing"
+)
+
+func TestProtocolV2Connect(t *testing.T) {
```
joao-r-reis, line 17: Remember that we talked about having these tests next to the protocol version negotiation ones

integration-tests/protocolv2_test.go
```diff
@@ -0,0 +1,171 @@
...
+	originAddress := "127.0.0.2"
+	targetAddress := "127.0.0.3"
+
+	serverConf := setup.NewTestConfig(originAddress, targetAddress)
+	proxyConf := setup.NewTestConfig(originAddress, targetAddress)
+	proxyConf.ControlConnMaxProtocolVersion = "3" // simulate protocol downgrade to V2
```
joao-r-reis, line 23: why this? if the server only supports v2 already then let's keep this as the default (which will be used by 99% of users) unless we want to actually test if this config setting works

integration-tests/setup/testcluster.go
```diff
@@ -452,6 +452,7 @@ func NewTestConfig(originHost string, targetHost string) *config.Config {
 	conf.ReadMode = config.ReadModePrimaryOnly
 	conf.SystemQueriesMode = config.SystemQueriesModeOrigin
 	conf.AsyncHandshakeTimeoutMs = 4000
+	conf.ControlConnMaxProtocolVersion = "3"
```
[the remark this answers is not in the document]
joao-r-reis: This is still unresolved

proxy/pkg/config/config.go
```diff
@@ -19,11 +20,12 @@ type Config struct {
...
-	ReadMode                string `default:"PRIMARY_ONLY" split_words:"true" yaml:"read_mode"`
-	ReplaceCqlFunctions     bool   `default:"false" split_words:"true" yaml:"replace_cql_functions"`
-	AsyncHandshakeTimeoutMs int    `default:"4000" split_words:"true" yaml:"async_handshake_timeout_ms"`
-	LogLevel                string `default:"INFO" split_words:"true" yaml:"log_level"`
+	PrimaryCluster                string `default:"ORIGIN" split_words:"true" yaml:"primary_cluster"`
+	ReadMode                      string `default:"PRIMARY_ONLY" split_words:"true" yaml:"read_mode"`
+	ReplaceCqlFunctions           bool   `default:"false" split_words:"true" yaml:"replace_cql_functions"`
+	AsyncHandshakeTimeoutMs       int    `default:"4000" split_words:"true" yaml:"async_handshake_timeout_ms"`
+	LogLevel                      string `default:"INFO" split_words:"true" yaml:"log_level"`
+	ControlConnMaxProtocolVersion string `default:"3" split_words:"true" yaml:"control_conn_max_protocol_version"` // Numeric Cassandra OSS protocol version or DseV1 / DseV2
```
joao-r-reis, line 28: We should put this as the highest version possible for the default because we will kind of infer the protocol version on the non control connections from the control connection versions.

proxy/pkg/zdmproxy/clienthandler.go
```diff
@@ -168,9 +168,12 @@ func NewClientHandler(
 	requestsDoneCtx, requestsDoneCancelFn := context.WithCancel(context.Background())
 
 	// Initialize stream id processors to manage the ids sent to the clusters
-	originFrameProcessor := newFrameProcessor(conf, nodeMetrics, ClusterConnectorTypeOrigin)
-	targetFrameProcessor := newFrameProcessor(conf, nodeMetrics, ClusterConnectorTypeTarget)
-	asyncFrameProcessor := newFrameProcessor(conf, nodeMetrics, ClusterConnectorTypeAsync)
+	originCCProtoVer := originControlConn.cqlConn.GetProtocolVersion().Load().(primitive.ProtocolVersion)
```
joao-r-reis, line 171: The `.GetProtocolVersion()` should do the `.Load().(primitive.ProtocolVersion` part instead of returning the atomic reference.

proxy/pkg/zdmproxy/clienthandler.go
```diff
@@ -2241,9 +2244,30 @@ func newFrameProcessor(conf *config.Config, nodeMetrics *metrics.NodeMetrics, co
...
 }
+
+// Calculate maximum number of stream IDs. Take the oldest protocol version negotiated between two clusters
+// and apply limit defined in proxy configuration. If origin or target cluster are still running protocol V2,
+// we will limit maximum number of stream IDs to 128 on both clusters. Logic is based on Java driver version 3.x.
+// Java driver 3.x was the last one supporting protocol V2. It establishes control connection first, and then
+// uses negotiated protocol version to configure maximum number of stream IDs on node connections. Driver does NOT
+// change the number of stream IDs on per node basis.
+func maxStreamIds(originProtoVer primitive.ProtocolVersion, targetProtoVer primitive.ProtocolVersion, conf *config.Config) int {
+	maxSupported := maxOutgoingPending
```
joao-r-reis, line 2261: Where is `maxOutgoingPending` being used? For the channel buffer size? We should keep the channel buffer sizes the same but add a new constant for the max number of stream ids in v3+ and use that constant here. Otherwise if a user tries to set 4000 max stream ids int he conf it won't work even though v3+ supports it.

integration-tests/protocolv2_test.go
```diff
@@ -0,0 +1,171 @@
...
+	require.Nil(t, err)
+}
+
+func TestProtocolV2Query(t *testing.T) {
+	originAddress := "127.0.0.2"
+	targetAddress := "127.0.0.3"
+
+	serverConf := setup.NewTestConfig(originAddress, targetAddress)
+	proxyConf := setup.NewTestConfig(originAddress, targetAddress)
+	proxyConf.ControlConnMaxProtocolVersion = "2"
```
joao-r-reis, line 57: same as above

integration-tests/protocolv2_test.go
```diff
@@ -0,0 +1,171 @@
...
+	require.Nil(t, err)
+
+	proxy, err := setup.NewProxyInstanceWithConfig(proxyConf) // starts the proxy
+	if proxy != nil {
+		defer proxy.Shutdown()
+	}
+	require.Nil(t, err)
+}
+
+func TestProtocolV2Query(t *testing.T) {
```
joao-r-reis, line 51: I'd say we can probably merge these 2 tests into a single one that connects and sends a query

integration-tests/connect_test.go
```diff
@@ -45,6 +46,80 @@ func TestGoCqlConnect(t *testing.T) {
...
+		{
+			name:                          "Cluster3.0_MaxCCProtoVer4_NegotiatedProtoVer4",
+			clusterVersion:                "3.0",
+			controlConnMaxProtocolVersion: "4",
+			negotiatedProtocolVersion:     primitive.ProtocolVersion4,
+		},
+		{
+			name:                          "Cluster4.0_MaxCCProtoVer4_NegotiatedProtoVer4",
+			clusterVersion:                "4.0",
+			controlConnMaxProtocolVersion: "4",
```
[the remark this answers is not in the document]
joao-r-reis: Offered some other suggestions for these tests  through slack

integration-tests/protocolv2_test.go
joao-r-reis: Offered some other suggestions for these tests  through slack

joao-r-reis (commented):
Some more comments

integration-tests/customhandler_test_utils.go
```diff
@@ -135,18 +153,32 @@ func systemLocalRow(cluster string, datacenter string, customPartitioner string,
 	if customPartitioner != "" {
 		partitionerValue = message.Column(customPartitioner)
 	}
+	if version >= primitive.ProtocolVersion3 {
```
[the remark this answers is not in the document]
lukasz-antoniak: That was based on manual test with C* 2.0.0 running (via CCM). I have tested now with 2.0.17 and I see `rpc_address` present.  Since this is just mock for test, I guess we should test most pessimistic approach.

We do processing of `system.local` query analogical to Java 3.x driver. ZDM tries to parse `rpc_address` column. If it is not there ZDM has no choice than use address that was given to establish control connection.
```
addr, port, err := ParseRpcAddress(false, row, defaultPort)
if addr == nil {
    // could not resolve address from system.local table (e.g. not present in C* 2.0.0)
    addr, port, err = ParseEndpoint(ccEndpoint)
}
```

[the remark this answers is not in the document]
joao-r-reis: Oh this just reminded me about https://github.com/datastax/zdm-proxy/issues/91 which is why we specifically say that we only support C* from 2.1.6 onwards (so we don't support 2.1.0-2.1.5).
If I understand correctly this means that now with this work we can close #91 AND we can now say that we support C* 2.1.0 and not just 2.1.6+ right?

README.md
```diff
@@ -95,12 +95,24 @@ containerized sandbox environment.
...
+ZDM Proxy requires origin and target clusters to have at least one protocol version in common. It is therefore not feasible
+to configure Apache Cassandra 2.0 as origin and 3.x / 4.x as target. Below table displays protocol versions supported by
+various C* versions:
+
+| Apache Cassandra | Protocol Version |
+|------------------|------------------|
+| 2.0              | V2               |
+| 2.1              | V2, V3           |
+| 2.2              | V3, V4           |
+| 3.x              | V3, V4           |
```
joao-r-reis, line 113: `2.2` supports V4? Are you sure? `3.x` doesn't support V2? Can you tell me the source for this? All this time I thought otherwise

README.md
```diff
@@ -110,8 +122,8 @@ migration process.
 
 In practice this means that ZDM Proxy supports the following cluster versions (as Origin and / or Target):
 
-- Apache Cassandra from 2.1+ up to (and including) Apache Cassandra 4.x. Apache Cassandra 2.0 support will be introduced 
-when protocol version v2 is supported.
+- Apache Cassandra from 2.1+ up to (and including) Apache Cassandra 4.x.
+- Apache Cassandra 2.0 up to 2.1.
```
joao-r-reis, line 126: Just edit the previous line to say `2.0+ up...` instead of `2.1+` and delete this one

integration-tests/customhandler_test_utils.go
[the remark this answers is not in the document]
lukasz-antoniak: I just used origin = 2.0.17 and target = 2.1.1. Connected Java driver 3.x, created tables (with collection columns), inserted and queried rows.

joao-r-reis (approved):
The tests look really good 👍

README.md
```diff
@@ -95,12 +95,24 @@ containerized sandbox environment.
...
+various C* versions:
+
+| Apache Cassandra | Protocol Version |
+|------------------|------------------|
+| 2.0              | V2               |
+| 2.1              | V2, V3           |
+| 2.2              | V2, V3, V4       |
+| 3.x              | V3, V4           |
+| 4.x              | V3, V4, V5       |
+
```
msmygit, line 115: With [OSS C* `5.0` around the corner](https://lists.apache.org/thread/1jtjt0h506y0kc4jt2v73jlgbm67mcb5), should we plan for it too (or) is it too ambitious?

lukasz-antoniak: V5 is definitely work for another PR. Moreover, as long as Astra and C* 5.x / DSE supports V4, they can still use ZDM for migration. Most CQL drivers do not support protocol V5 yet.

PR #123, opened 2024-07-01
Support providing configuration from YAML files

lukasz-antoniak:


proxy/pkg/config/config.go
```diff
@@ -133,12 +135,46 @@ func New() *Config {
...
+
+func (c *Config) LoadConfig() (*Config, error) {
+	err := c.parseEnvVars()
+	if err != nil {
+		return nil, err
+	}
+
+	err = c.loadFromFiles()
+	if err != nil {
+		return nil, err
```
joao-r-reis, line 177: I don't understand this, what's the goal of parsing both at the same time, shouldn't we parse just the CONFIG_FILES env var first to see if it's set and if it's set we just skip parsing all other env vars and just parse the config files?

proxy/pkg/config/config.go
```diff
@@ -133,12 +135,46 @@ func New() *Config {
...
+			continue
+		}
+		file, err := os.Open(path)
+		if err != nil {
+			return fmt.Errorf("could not read configuration file %v: %w", path, err)
+		}
+		defer file.Close()
+
+		dec := yaml.NewDecoder(file)
+		if err = dec.Decode(c); err != nil {
```
joao-r-reis, line 151: what happens if the same config settings are defined in multiple files? As it stands it seems that they will be overwritten which is not a very clear behavior. We should probably return an error and fail the launch if that happens.

lukasz-antoniak: It was intentional to allow to override. At the end we already print all applied configuration variables, so there should be no confusion.

joao-r-reis: I heavily disagree with this, I don't see any benefit to it and it might just lead to user error, users never check the configuration that is printed in the logs, it's mostly for us devs to troubleshoot issues.

alicel: I hadn't thought about the possibility of the same value being set in multiple files, but it is a good point and definitely a risk.

I would see the value in supporting multiple files so that users can have one with stricter governance and another with easier access. For example, they may want to have one file containing credentials / SCB path / contact points, with restricted access, and another with things like read mode, primary cluster, log level etc etc which can be more easily accessed.

This could also be managed by having clearly structured and named files, from which we extract only the expected properties. This would not leave it up to the user to define any property in any file, and would ensure that each property is defined and considered exactly once. The downside is that we need to parse each file specifically for the expected content, but the configuration properties themselves do not tend to change, so I wouldn't see it as a big problem.

joao-r-reis: I wonder if we are overengeneering this a bit... In general most tools just allow you to provide a config file and that's it. In the case of DSE, DSE allows you to set a config setting that is a path to a key file so that DSE can decrypt the credentials.

Can we take a step back and actually think if we need multiple config files? I feel like the DSE approach is a simple one but effective no?

proxy/pkg/config/config.go
```diff
@@ -133,12 +136,42 @@ func New() *Config {
...
 }
 
+func (c *Config) loadFromFile(configFile string) error {
+	file, err := os.Open(configFile)
+	if err != nil {
+		return fmt.Errorf("could not read configuration file %v: %w", configFile, err)
+	}
+	defer file.Close()
+
+	def.SetDefaults(c) // apply default tag, it is not supported by YAML decoder
```
joao-r-reis, line 146: nice idea 👍

proxy/pkg/config/config.go
```diff
@@ -148,7 +181,7 @@ func (c *Config) ParseEnvVars() (*Config, error) {
 
 	log.Infof("Parsed configuration: %v", c)
 
-	return c, nil
+	return c, err
```
joao-r-reis, line 184: I'd prefer to keep `nil` here since it is more clear that there will always never be an err at this point

proxy/main_profiling.go
```diff
@@ -61,5 +62,5 @@ func main() {
 		}()
 	}
 
-	launchProxy(true)
+	launchProxy(true, configFile)
```
joao-r-reis, line 65: did you try to compile this? I don't think this builds. I think we should put the configfile flag on a separate file that always builds regardless of whether profiling mode is enabled or not.

proxy/main.go
```diff
@@ -16,7 +16,8 @@ import (
 // TODO: to be managed externally
 const ZdmVersionString = "2.2.0"
 
-var displayVersion = flag.Bool("version", false, "Display the ZDM proxy version and exit")
+var displayVersion = flag.Bool("version", false, "display the ZDM proxy version and exit")
```
joao-r-reis, line 19: We should move all the code on this file to a separate file that always builds regardless of whether profiling mode is enabled or not. And the config file flag should be moved to that file as well.

The only difference between `main_profiling.go` and `main.go` should be the profiling stuff. At the moment `displayVersion` is not enabled during profile mode which makes no sense.

I know this wasn't done on this PR but since you're adding a new configfile flag we should probably get this fixed now.

joao-r-reis (commented):
Looks good, added a few comments.

We should add a "reference" yaml file so that we can point users when they ask about its structure. It wouldn't be used by the proxy like the java driver does it but it would sit on the repo so that we can use it for documentation and example purposes. It should contain all settings (except the performance tuning ones) including the default values (or commented if they don't have a default value)

proxy/main.go
lukasz-antoniak: I did not want to add new small file and moved the code to `launch.go`. I refactored also version displaying not to duplicate code. See if it is OK for you.

joao-r-reis (approved):

docs/assets/zdm-config-reference.yml
```diff
@@ -0,0 +1,164 @@
...
+# reads will not be sent. This has no impact on the handling of synchronous requests: the ZDM Proxy will
+# continue to handle all synchronous reads and writes normally.
+async_handshake_timeout_ms: 4000
+
+# Specifies logging level.
+log_level: INFO
+
+# TODO: ProxyTopologyIndex
+# TODO: ProxyTopologyAddresses
+# TODO: ProxyTopologyNumTokens
```
joao-r-reis, line 30: you want to work on these 3 later? I'm ok with merging as is

docs/assets/zdm-config-reference.yml
```diff
@@ -0,0 +1,164 @@
...
+heartbeat_interval_ms: 30000
+
+# Below properties define reconnection strategy for establishing control connection.
+heartbeat_retry_interval_min_ms: 250
+heartbeat_retry_interval_max_ms: 30000
+heartbeat_retry_backoff_factor: 2
+
+# Control connection failure threshold. If threshold is exceeded,
+# readiness probe of ZDM will report failure and pod will be recreated.
+heartbeat_failure_threshold: 1
```
joao-r-reis, line 164: Maybe we can comment out all of the optional settings (but have the commented line here anyway just so that it is easy for someone to uncomment and set a specific value).

The required settings should be left uncommented like they are now 👍

docs/assets/zdm-config-reference.yml
```diff
@@ -0,0 +1,164 @@
...
+# origin_tls_server_ca_path:
+
+# Public key used when establishing connectivity with origin cluster.
+# origin_tls_client_cert_path:
+
+# Private key used to secure communication with origin cluster.
+# origin_tls_client_key_path:
+
+# Comma separated ist of target cluster contact points.
+target_contact_points: target.datastax.com
```
joao-r-reis, line 63: maybe 127.0.0.2 as target and 127.0.0.1 as origin? feels a bit weird to see a valid hostname here (even if it doesn't exist, at least I don't think it does? but the datastax.com domain definitely exists 😅 )

joao-r-reis (approved):
the reference yml looks really good 💯

docs/assets/zdm-config-reference.yml
lukasz-antoniak: Good point, fixed.

docs/assets/zdm-config-reference.yml
lukasz-antoniak: I left all required and sensitive properties uncommented, but commented majority.

docs/assets/zdm-config-reference.yml
```diff
@@ -0,0 +1,170 @@
...
+# Timeout (in ms) when performing the initialization (handshake) of a proxy-to-secondary cluster
+# connection that will be used solely for asynchronous dual reads. If this timeout occurs, the asynchronous
+# reads will not be sent. This has no impact on the handling of synchronous requests: the ZDM Proxy will
+# continue to handle all synchronous reads and writes normally.
+# async_handshake_timeout_ms: 4000
+
+# Specifies logging level.
+# log_level: INFO
+
+# List of peer ZDM proxy instances. This configuration parameter should be identical through all ZDM proxies.
```
joao-r-reis, line 28: Should stress here that the order matters and even the order should be the same on all proxies

docs/assets/zdm-config-reference.yml
```diff
@@ -0,0 +1,170 @@
...
+# proxy_topology_addresses: 127.0.1.1, 127.0.1.2, 127.0.1.3
+
+# Index of local ZDM proxy instance within "proxy_topology_addresses" list.
```
joao-r-reis, line 31: Maybe a small example like the one I gave you in slack would be good to illustrate this. `0` on the `127.0.1.1`, `1` on the `127.0.1.2` and `2` on the `127.0.1.3`

docs/assets/zdm-config-reference.yml
```diff
@@ -0,0 +1,170 @@
...
+# proxy_topology_index: 0
+
+# Number of tokens each proxy isntance owns. The default value of 8 should workf for majority of use-case.
```
joao-r-reis, line 34: typo on `workf` and should be "majority of use cases".

docs/assets/zdm-config-reference.yml
```diff
@@ -0,0 +1,170 @@
+# This variable determines which cluster is currently considered the primary cluster.
```
joao-r-reis, line 1: Add a paragraph at the top saying that required settings are uncommented and all optional ones are commented.

joao-r-reis (approved):

PR #125, opened 2024-07-11
Expose read and write request latency in metrics

lukasz-antoniak:


proxy/pkg/metrics/node_metrics.go
```diff
@@ -280,7 +280,7 @@ type NodeMetricsInstance struct {
 	UnavailableErrors Counter
 	OtherErrors       Counter
 
-	RequestDuration Histogram
+	RequestDuration map[string]Histogram
```
joao-r-reis, line 283: Why not follow what we have on `ProxyMetrics`? We already have write and read metrics on the `ProxyMetrics` , just need to replicate what's there.

proxy/pkg/metrics/node_metrics.go
```diff
@@ -298,15 +298,31 @@ func CreateCounterNodeMetric(metricFactory MetricFactory, nodeDescription string
...
 	m, err := metricFactory.GetOrCreateHistogram(
-		mn.WithLabels(map[string]string{nodeLabel: nodeDescription}), buckets)
+		mn.WithLabels(map[string]string{nodeLabel: nodeDescription}).WithLabels(customLabels), buckets)
 	if err != nil {
 		return nil, err
 	}
 	return m, nil
 }
 
+func CreateHistogramNodeRequestDurationMetrics(metricFactory MetricFactory, nodeDescription string, mn Metric, buckets []float64) (map[string]Histogram, error) {
```
joao-r-reis, line 314: Can you try replicating what we have on the `ProxyMetrics` side, i.e., splitting `OriginRequestDuration` into `OriginWritesDuration` and `OriginReadsDuration` (same for the target one)? Shouldn't that work and remove the need to have this new code here on these 2 functions?

proxy/pkg/metrics/proxy_metrics.go
```diff
@@ -18,14 +19,18 @@ const (
 	failedWritesFailedOnClusterTypeLabel = "failed_on"
 
 	requestDurationName        = "proxy_request_duration_seconds"
-	requestDurationTypeLabel   = "type"
+	RequestDurationTypeLabel   = "type"
```
joao-r-reis, line 22: why export these 3 constants

joao-r-reis (commented):
Some comments. This approach probably works fine but I think we should do the same thing as we are doing for the read/write metrics on the proxy metrics to be consistent.

proxy/pkg/metrics/node_metrics.go
```diff
@@ -298,9 +299,13 @@ func CreateCounterNodeMetric(metricFactory MetricFactory, nodeDescription string
 	return m, nil
 }
 
-func CreateHistogramNodeMetric(metricFactory MetricFactory, nodeDescription string, mn Metric, buckets []float64) (Histogram, error) {
+func CreateHistogramNodeMetric(metricFactory MetricFactory, nodeDescription string, mn Metric, buckets []float64, labels ...string) (Histogram, error) {
```
joao-r-reis, line 302: the `labels ...string` is not very obvious, it's a list where evens are keys and odds are values for those keys... Can we make it a simple `map[string]string` instead? Or create a new label type:

1: 
```
type MetricLabel struct {
   Key string
   Value string
}
func CreateHistogramNodeMetric(metricFactory MetricFactory, nodeDescription string, mn Metric, buckets []float64, labels ...MetricLabel) (Histogram, error) { }

metrics.CreateHistogramNodeMetric(metricFactory, asyncNodeDescription, metrics.AsyncRequestDuration, asyncBuckets, metrics.MetricLabel{Key: metrics.RequestDurationTypeLabel, Value: metrics.TypeReads})
```
2:
```
func CreateHistogramNodeMetric(metricFactory MetricFactory, nodeDescription string, mn Metric, buckets []float64, labels map[string]string) (Histogram, error) { }

metrics.CreateHistogramNodeMetric(metricFactory, asyncNodeDescription, metrics.AsyncRequestDuration, asyncBuckets, map[string]string{metrics.RequestDurationTypeLabel: metrics.TypeReads})
```

joao-r-reis (approved):
looks good, left just a NIT comment, feel free to ignore it if you think it's not important

PR #139, opened 2025-02-25
Ignore forwarding CQL requests for Insights Client

lukasz-antoniak:


proxy/pkg/zdmproxy/cqlparser.go
```diff
@@ -213,13 +214,18 @@ func getRequestInfoFromQueryInfo(
 		}
 	} else if queryInfo.getStatementType() == statementTypeUse {
 		sendAlsoToAsync = true
+	} else if queryInfo.getStatementType() == statementTypeCall {
+		// RPC client, typically used by "insights reporter" for DSE
```
joao-r-reis, line 218: I wonder if it's safer to parse the actual function being called, I'm not sure if there are other RPCs being used with DSE, maybe stuff related to Spark and/or search

joao-r-reis: Storing the first identifier would be enough I think? And then check if that identifier matches `InsightsRpc` (case insensitive to be safe)

lukasz-antoniak: Very good idea. Updated code.

joao-r-reis (approved):

joao-r-reis:
Very nice stuff 👍

PR #149, opened 2025-06-09
Log CQL request ID for distributed tracing

lukasz-antoniak:
Fixes https://github.com/datastax/zdm-proxy/issues/150.


joao-r-reis (commented):
The tests are really good, they helped me understand the feature better.

I added a comment about how you plugged in the code related to protocol level frame modification. I think we need some changes for the code to be more future proof, it is very clear that the code in `main` atm wasn't written taking into account potential protocol level frame modifications like what we want to do now with the custom payload. As a result the code related to CQL query string modification is not easily reused if we want to add another layer of abstraction on top so that there can be CQL query modification in addition to frame modification. I added a comment with my proposal

proxy/pkg/zdmproxy/querymodifier.go
```diff
@@ -48,14 +83,10 @@ func (recv *QueryModifier) replaceQueryString(currentKeyspace string, context *f
...
-		return nil, nil, fmt.Errorf("could not replace query string in request '%v': %w", requestType, err)
+		return false, nil, nil, fmt.Errorf("could not replace query string in request '%v': %w", requestType, err)
 	}
 
-	newRawFrame, err := defaultCodec.ConvertToRawFrame(newFrame)
-	if err != nil {
-		return nil, nil, fmt.Errorf("could not convert modified frame to raw frame: %w", err)
-	}
-	return NewInitializedFrameDecodeContext(newRawFrame, newFrame, newStatementsQueryData), replacedTerms, nil
+	return true, NewInitializedFrameDecodeContext(context.frame, newFrame, newStatementsQueryData), replacedTerms, nil
```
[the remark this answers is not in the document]
joao-r-reis: I should add that the goal of having a type like `FrameDecodeContext` is so that we can delay the decoding of the frame until where it is really needed (if it is needed at all) and when that happens then future calls will have access to the decoded frame so there's no duplicate decoding of the same frame. The "cql string" inspection was added to that type but maybe it shouldn't have been, I'm not even sure if the code tries to "inspect" the cql query string anywhere else other than the client handler

proxy/pkg/zdmproxy/clienthandler.go
```diff
@@ -1365,18 +1365,26 @@ func (ch *ClientHandler) forwardRequest(request *frame.RawFrame, customResponseC
...
 		ch.forwardSystemQueriesToTarget, ch.topologyConfig.VirtualizationEnabled, ch.forwardAuthToTarget, ch.timeUuidGenerator)
+
+	switch requestInfo.(type) {
+	case *InterceptedRequestInfo:
+		// do not log system queries
+		break
+	default:
+		reqId := context.GetRequestId(ch.conf)
+		if reqId != nil {
+			log.Infof("Request id %v (hex) for query with stream id %d and %v", hex.EncodeToString(reqId), request.Header.StreamId, request.Header.OpCode)
```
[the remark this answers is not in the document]
joao-r-reis: With rate limiting and this being behind a flag that is opt in I think INFO is ok

proxy/pkg/zdmproxy/querymodifier.go
```diff
@@ -163,3 +154,41 @@ func queryOrPrepareRequiresQueryReplacement(statementsQueryData []*statementQuer
 
 	return requiresQueryReplacement(statementsQueryData[0]), statementsQueryData[0], nil
 }
+
+func (recv *QueryModifier) assignRequestId(protoVer primitive.ProtocolVersion, decodedFrame *frame.Frame) (bool, error) {
+	if protoVer < primitive.ProtocolVersion4 {
+		return false, nil
```
joao-r-reis, line 160: Is it not possible to implement this in v3 or v2?

proxy/pkg/zdmproxy/querymodifier.go
```diff
@@ -163,3 +154,41 @@ func queryOrPrepareRequiresQueryReplacement(statementsQueryData []*statementQuer
...
+	}
+	op := decodedFrame.Header.OpCode
+	if op != primitive.OpCodePrepare && op != primitive.OpCodeExecute && op != primitive.OpCodeQuery && op != primitive.OpCodeBatch {
+		return false, nil
+	}
+
+	customPayload := decodedFrame.Body.CustomPayload
+	allow := recv.rateLimiters.Allow(RequestIdTracingLimit)
+	if !allow {
+		// remove ID potentially provided by the upstream application
```
joao-r-reis, line 170: We only enter this if tracing is enabled in the configuration so if it is not enabled the app provided request ID will remain on the frame and `.GetRequestId()` will still return a valid request id which means the request id will still be logged.

proxy/pkg/zdmproxy/clienthandler.go
```diff
@@ -1403,6 +1413,54 @@ func (ch *ClientHandler) forwardRequest(request *frame.RawFrame, customResponseC
...
+		// re-encode the frame if original request has been modified
+		newRawFrame, err := defaultCodec.ConvertToRawFrame(decodedFrame)
+		if err != nil {
+			return nil, nil, fmt.Errorf("could not convert modified frame to raw frame: %w", err)
+		}
+		return NewInitializedFrameDecodeContext(newRawFrame, decodedFrame, statementsQueryData), replacedTerms, nil
+	}
+
+	return context, replacedTerms, nil
+}
```
joao-r-reis, line 1462: I like how this function turned out, great stuff 💯

proxy/pkg/zdmproxy/clienthandler.go
```diff
@@ -1403,6 +1413,61 @@ func (ch *ClientHandler) forwardRequest(request *frame.RawFrame, customResponseC
...
+	var modifiedFrame, modifiedQuery bool
+
+	// replace runtime-evaluated CQL functions in queries
+	if ch.conf.ReplaceCqlFunctions {
+		decodedFrame, statementsQueryData, err = context.GetOrDecodeAndInspect(currentKeyspace, ch.timeUuidGenerator)
+		if err != nil {
+			if !errors.Is(err, NotInspectableErr) {
+				return nil, nil, fmt.Errorf("could not check whether query needs replacement for a '%v' request: %w",
+					decodedFrame.Header.OpCode.String(), err)
+			}
```
joao-r-reis, line 1434: missing something here to handle the `else` scenario where `err` is still `nil` and therefore `decodedFrame` is `nil`

proxy/pkg/zdmproxy/querymodifier.go
lukasz-antoniak: Fixed.

PR #155, opened 2025-08-26
[DRAFT] Update gocql to 2.0

joao-r-reis:
Marked as DRAFT until my gocql PRs are merged so that we can use gocql directly and not my fork


joao-r-reis:
Handled in #157

PR #160, opened 2026-01-30
Fix decompression buffer size for LZ4

lukasz-antoniak:
Upgrade `go-cassandra-native-protocol` library to after fixing https://github.com/datastax/go-cassandra-native-protocol/pull/61.


joao-r-reis (approved):
👍

Issue #177, opened 2026-09-18
arm64 docker images


Tolsto:
Please release the images on Dockerhub as multi-arch including arm64

joao-r-reis:
Note that this project has migrated to Quay.io as a replacement of Dockerhub (`docker pull quay.io/datastax/zdm-proxy`)

PR #178, opened 2026-09-18
Enable multi-arch builds

Tolsto:
Prerequisite for https://github.com/datastax/zdm-proxy/issues/177


joao-r-reis:
#179 has the full implementation to address #177

